PR #51, opened 2020-06-30
API Updates

tyler-whitman:
Added Elections and Account Stats

### DIFF
--- a/packages/http/src/models/Account.ts
+++ b/packages/http/src/models/Account.ts
@@ -22,6 +22,29 @@ function toBalance(value: number | undefined, type: CurrencyType): Balance | und
   return new Balance(value, type)
 }
 
+export interface HTTPStatsObject {
+  last_week: Array<TimelineStats>
+  last_month: Array<TimelineStats>
+  last_day: Array<TimelineStats>
+}
+
+export interface TimelineStats {
+  timestamp: string
+  balance: number
+}
+
+export class AccountStats {
+  constructor(data: HTTPStatsObject) {
+    this.week = data.last_week.reverse()
+    this.month = data.last_month.reverse()
+    this.day = data.last_day.reverse()
+  }
+
+  public week: Array<TimelineStats>
+  public month: Array<TimelineStats>
+  public day: Array<TimelineStats>
+}
+
 export default class Account {
   private client: Client
   public speculativeNonce?: number
--- a/packages/http/src/resources/Elections.ts
+++ b/packages/http/src/resources/Elections.ts
@@ -1,0 +1,28 @@
+import type Client from '../Client'
+import ResourceList from '../ResourceList'
+import Election from '../models/Election'
+
+interface ListParams {
+  cursor?: string
+}
+
+export default class Elections {
+  private client!: Client
+
+  constructor(client: Client) {
+    this.client = client
+  }
+
+  async list(params: ListParams = {}): Promise<ResourceList<Election>> {
+    const url = '/elections'
+    const response = await this.client.get(url, { cursor: params.cursor })
+    const { data: { data: elections, cursor } } = response
+    return new ResourceList(elections, this.list.bind(this), cursor)
+  }
+
+  async get(hash: string): Promise<Election> {
+    const url = `/transactions/${hash}`
+    const { data: { data: election } } = await this.client.get(url)
+    return election
+  }
+}
--- a/packages/http/src/resources/Elections.ts
+++ b/packages/http/src/resources/Elections.ts
@@ -21,6 +21,7 @@ export default class Elections {
   }
 
   async get(hash: string): Promise<Election> {
+    // using transactions api for lookup, there is no /elections/hash endpoint
     const url = `/transactions/${hash}`
     const { data: { data: election } } = await this.client.get(url)
     return election
--- a/packages/http/src/resources/__tests__/Elections.spec.ts
+++ b/packages/http/src/resources/__tests__/Elections.spec.ts
@@ -1,7 +1,7 @@
 import nock from 'nock'
 import Client from '../../Client'
 
-const challengeFixture = (params = {}) => ({
+const ellectionsFixture = (params = {}) => ({
   type: 'consensus_group_v1',
   time: 1234567890,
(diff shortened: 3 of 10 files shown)


packages/http/src/models/Account.ts
```diff
@@ -22,6 +22,29 @@ function toBalance(value: number | undefined, type: CurrencyType): Balance | und
...
+  last_month: Array<TimelineStats>
+  last_day: Array<TimelineStats>
+}
+
+export interface TimelineStats {
+  timestamp: string
+  balance: number
+}
+
+export class AccountStats {
```
allenan, line 36: should this be in a separate file maybe?

packages/http/src/resources/__tests__/Elections.spec.ts
```diff
@@ -0,0 +1,52 @@
...
...sLN6LdoskQr2XXkxwoJR4jM7eOKpG1PJ0XPb7snVhxn7BB4y3bPCE3eG9d4-ZXkUTPRZKaTl1ssdSK6aRfjGybAQTDGv6CXGcZPUbMbWi-RDxSK0Tl8li1Nq1LvtNoklJ7BVC-cPhPOY373VZ6y2SAHNGrtfNzQjmWyndTVbOGLYRmjW3t3O7LR4w_mshtyZ09Imvm3F4jM_DyMUtfcYuz0p6LxfvWEeNcbwPB7eJpW6WkcKH1sRFJsNkRRDMltNWs38LPyXv_5Ur4MjuCXbAo2Ur-bsuG-QsMdM3btj67BGcZIQ1ntE5TiVI6dr-c6806VHAeV8CguunlFZP757HjaoloETsf92qjBHPgNUG5Npjj_IGr4IBI-TPLXFZvo6w7LFwI0VcxKLVQIjs5WyS3EGJ-69065KU9BpK9U6o6JCy28XrWKnMi7FfsmyiPBh1LLZ1e7X9kDkl571VgFCxGZHxPZlxdZTXtK1ht5IU51JL_wjd63DmjzSWdf3jYUli421-V8H65OH6fszapMuNOi5AgePGhDojaHGYC7mA6MZyk_H1GSgx_v3QKqLcTiYJ2rhVLZHJ4An1CACEwpWHM-F9MPhFhbPMm5wFK035Utk9-EIAk8qCDiFn7CW5vbW8j8vCYrNrOT9siY08Z1y1cjEia8pGEarfcTyVf5rz9W415z5U5N_jSnA40kQyAVlmnQSufrnrIdMhHpk9Kmf3fmWmtyvb2DjwIhh3xUKLUVK-kIc71zt9DeV5cWlcqXow3cGIzhhqW4yFGxCXhJvyBEJSJf_I5usvPw4J_OfGklm6o55T8DbWgtAWf-fLgsF6-FGE0StgjcYmyWk79wSWWhY8LW3m782KNTTwZuw0WTMsasl0qA5OBqzm0joO1rTtA63qSzygdFXpkOnjeXOCpG6_8iEurenhxSMTXRnozEXBRcWy196Yk9G7e__4GipfWHnR9RTGe7uVX54SRMZq9UmElcXVjOCK5tc2oZ8stNb6qBDQEIOfsF7HdMvrb54fk-A5mbtqCHT4saKikfiasw74FIllqKmARx50ppN1TkS4vL7g6G5lwG9_piS13G_v56RkkjfK3n0s9Obhydc6anLfU58vkkcMPxkXMypZ0cetsskf4JdvLsrLdo0rTsuczZrBh3U-kyFSoIWC-ciDyTPBhJQMotvnGEMRz8P7qiJDc',
```
allenan, line 7: for the mock, unless this proof means something, we could just shorten this to `fake-proof` or something

packages/http/src/resources/__tests__/Elections.spec.ts
```diff
@@ -0,0 +1,52 @@
+import nock from 'nock'
+import Client from '../../Client'
+
+const ellectionsFixture = (params = {}) => ({
```
allenan, line 4: sp.

packages/http/src/models/Account.ts
tyler-whitman: Because the endpoint is `/accounts/id/stats` I added it under Accounts. So you call `client.accounts.getStats('my-address')` thats why I put it in Account.
I can also move it out to its own object, maybe `client.accountStats.get('my-address')`. Let me know what you think

packages/http/src/resources/__tests__/Elections.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  ],
+  height: 123456,
+  hash: 'fake-hash',
+  delay: 0,
+  ...params,
+})
+
+describe('get', () => {
+  nock('https://api.helium.io')
+    .get('/v1/transactions/fake-hash')
```
allenan, line 23: should this be mocking the elections endpoint rather than transactions?

tyler-whitman: So there is no `/v1/elections/fake-hash` they are just transactions. So the client.elections.get actually calls /transactions

packages/http/src/resources/Elections.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  async list(params: ListParams = {}): Promise<ResourceList<Election>> {
+    const url = '/elections'
+    const response = await this.client.get(url, { cursor: params.cursor })
+    const { data: { data: elections, cursor } } = response
+    return new ResourceList(elections, this.list.bind(this), cursor)
+  }
+
+  async get(hash: string): Promise<Election> {
+    const url = `/transactions/${hash}`
```
allenan, line 24: this should also probably be elections rather than transactions

packages/http/src/resources/__tests__/Elections.spec.ts
tyler-whitman: Fixed

packages/http/src/resources/Elections.ts
tyler-whitman: Mentioned in comment above but that path doesnt exist on the API

packages/http/src/resources/__tests__/Elections.spec.ts
tyler-whitman: Updated

packages/http/src/resources/__tests__/Elections.spec.ts
allenan: hm i see, i thought it might have been a typo hiding behind a bad mock. that's fine then. maybe add a comment so we remember?

packages/http/src/models/Account.ts
allenan: i'm fine with the call being nested under `client.accounts.getStats`. I was just thinking since it's a class, maybe we'd want it in its own file to make searching for it easier, but i don't feel too strongly about it

packages/http/src/resources/__tests__/Elections.spec.ts
tyler-whitman: Yeah let me add a comment

PR #52, opened 2020-06-30
add floatBalance to Balance and update account stats

tyler-whitman:


packages/http/src/models/Balance.ts
```diff
@@ -17,12 +17,14 @@ BigNumber.config({
...
+  public floatBalance: number
   public type: CurrencyType
   private bigBalance: BigNumber
 
   constructor(integerBalance: number | undefined, type: CurrencyType) {
     this.integerBalance = integerBalance || 0
     this.bigBalance = new BigNumber(this.integerBalance)
+    this.floatBalance = this.bigBalance.times(type.coefficient).toNumber()
     this.type = type
   }
```
allenan, line 30: maybe we can use `floatBalance` in `toString` in place of `const number = ...`

tyler-whitman: Good call let me update

PR #60, opened 2020-07-10
support for ecc_compact, verification of supported keytypes

allenan:
adds support for ecc_compact key types and adds additional verification that a b58 address is derived from a key type we support.

### DIFF
--- a/packages/crypto-react-native/src/utils.ts
+++ b/packages/crypto-react-native/src/utils.ts
@@ -67,3 +67,15 @@ export const bs58ToBin = (bs58Address: string): Buffer => {
 
   return payload
 }
+
+export const bs58KeyType = (bs58Address: string): number => {
+  const bin = bs58ToBin(bs58Address)
+  const keyType = Buffer.from(bin).slice(0, 1)[0]
+  return keyType
+}
+
+export const bs58Version = (bs58Address: string): number => {
+  const bin = bs58.decode(bs58Address)
+  const version = bin.slice(0, 1)[0]
+  return version
+}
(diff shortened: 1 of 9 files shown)


packages/crypto-react-native/src/utils.ts
```diff
@@ -67,3 +67,15 @@ export const bs58ToBin = (bs58Address: string): Buffer => {
 
   return payload
 }
+
+export const bs58KeyType = (bs58Address: string): number => {
+  const bin = bs58ToBin(bs58Address)
```
tyler-whitman, line 72: Do these throw if the  bs58Address string is invalid?

packages/crypto/src/KeyType.ts
```diff
@@ -0,0 +1,9 @@
+export const ECC_COMPACT_KEY_TYPE = 0
```
tyler-whitman, line 1: Could we share these files between crypto and crypto-react-native?

tyler-whitman:
Looks good to me

packages/crypto/src/KeyType.ts
allenan: I'd like to but there are some subtle differences because react-native uses a different sodium library. I'd like to maybe refactor the common stuff out into a separate package and then have the `crypto` packages just export a consistent api for the sodium packages. Just haven't gotten around to doing that yet

packages/crypto-react-native/src/utils.ts
allenan: Yes, the bs58ToBin function throws in 2 places, 1) if it cannot decode bs58 and 2) if the checksum is invalid

PR #64, opened 2020-07-17
Add getPriceAtBlock and getPredictedPrice

AddressXception:
Add `getPriceAtBlock` and `getPredictedPrice` methods

## To Test:
1. `yarn test`
2. expected: all tests pass

### DIFF
--- a/packages/http/src/resources/Oracle.ts
+++ b/packages/http/src/resources/Oracle.ts
@@ -10,9 +10,27 @@ export default class Oracle {
   }
 
   async getCurrentPrice(): Promise<any> {
-    const url = `/oracle/prices/current`
-    const { data: { data: { price } } } = await this.client.get(url)
-    
-    return new Balance(price, CurrencyType.usd)
+    const url = '/oracle/prices/current'
+    const { data: { data: { price, block } } } = await this.client.get(url)
+
+    return { price: new Balance(price, CurrencyType.usd), height: block }
+  }
+
+  async getPriceAtBlock(height: number): Promise<any> {
+    const url = `/oracle/prices/${height}`
+    const { data: { data: { price, block } } } = await this.client.get(url)
+
+    return { price: new Balance(price, CurrencyType.usd), height: block }
+  }
+
+  async getPredictedPrice(): Promise<any> {
+    const url = '/oracle/predictions'
+    const response = await this.client.get(url)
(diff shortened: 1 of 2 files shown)


packages/http/src/models/Balance.ts
```diff
@@ -22,12 +22,14 @@ export default class Balance {
   public floatBalance: number
   public type: CurrencyType
   public bigBalance: BigNumber
+  public height?: number
 
-  constructor(integerBalance: number | undefined, type: CurrencyType) {
+  constructor(integerBalance: number | undefined, type: CurrencyType, height?: number) {
```
AddressXception, line 27: i wasn't sure the preferred order here so i just put it at the end

packages/http/src/resources/Oracle.ts
```diff
@@ -10,9 +10,27 @@ export default class Oracle {
...
+    const { data: { data: { price, block } } } = await this.client.get(url)
+
+    return new Balance(price, CurrencyType.usd, block)
+  }
+
+  async getPredictedPrice(): Promise<any> {
+    const url = '/oracle/predictions'
+    const response = await this.client.get(url)
+    if (!response.data.data || !response.data.data.length) {
+      return []
```
AddressXception, line 30: to be consistent with the way the API behaves, returning an empty array.  Could be changed.

allenan: I think that's fine to stick with how the API behaves

packages/http/src/models/Balance.ts
allenan: Is it necessary to include `height` within Balance? I know the API returns `time` and `height` depending on the route so there's already a bit of inconsistency there. Maybe we could do like you do in the predicted price route and return a `{ price, height }` object?

allenan: That would also have the advantage of conforming more closely with what the API returns

AddressXception: i like that, i'll make the change there

allenan:
this looks good, i'm just going to revert out the changes in Balance so that it doesn't create a diff

allenan:
actually I might need you to make that change @AddressXception. Could you revert out the comma changes in Balance.ts so that it doesn't create unnecessary diffs? Then I'll get this merged and published for you

PR #76, opened 2020-08-11
TokenBurnV1

tyler-whitman:


packages/transactions/README.md
```diff
@@ -1,9 +1,18 @@
...
+    payee: 'payee_address',
+    amount: 10,
+    nonce: 1,
+    memo: 1234567890,
+})
 
-```
-// TODO: DEMONSTRATE API
+// sign transaction
+const signedTransaction = await transaction.sign({ payer: 'payer_address' })
```
allenan, line 17: i think the argument to `payer` here would be the full keypair

tyler-whitman: Ah let me fix it

PR #77, opened 2020-08-12
pendingTransactions.get now returns an array

tyler-whitman:
In the case that a transaction fails and is submitted again `pendingTransactions.get` will return multiple pending transactions.

### DIFF
--- a/packages/http/src/resources/PendingTransactions.ts
+++ b/packages/http/src/resources/PendingTransactions.ts
@@ -7,6 +7,7 @@ type Context = Account
 
 export default class PendingTransactions {
   private client!: Client
+
   private context?: Context
 
   constructor(client: Client, context?: Context) {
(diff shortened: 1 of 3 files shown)


packages/http/src/resources/PendingTransactions.ts
```diff
@@ -7,17 +7,22 @@ type Context = Account
...
-  async get(hash: string): Promise<PendingTransaction> {
+  async get(hash: string): Promise<PendingTransaction[]> {
     const url = `/pending_transactions/${hash}`
     const { data: { data } } = await this.client.get(url)
-    return new PendingTransaction(data)
+    const pendingTransactions: PendingTransaction[] = []
+    data.forEach((txn: HTTPPendingTransactionObject) => {
+      pendingTransactions.push(new PendingTransaction(txn))
+    })
+    return pendingTransactions
```
allenan, line 25: should this return a resource list instead of an array? not sure if they're ever going to be cursor-paginated, but at least it would be consistent with all the other methods that return lists of items. and you could call `take` on it etc

tyler-whitman: The reason I didnt do that is because it wasnt cursor paginated, so its always returning the entire array. Do you think we need resource list for that?

tyler-whitman: The list should be fairly small, as it really only have multiple when a transactions fails and hopefully we aren't failing more than like 10 times haha

allenan: I think we should still return it as a resource list even if it's not paginated now because those still act like an array, and also have the `take` helper. that way we're always returning a consistent resource

tyler-whitman: Ok let me update it

tyler-whitman: Updated now

PR #88, opened 2020-10-02
list hotspot witnesses

tyler-whitman:
Implement https://developer.helium.com/blockchain/api/api-hotspots#witnesses-for-a-hotspot


allenan (approved):
looks good!

PR #96, opened 2020-10-18
refactor(crypto): sha256, hash, and webpack

### LINKED ISSUE #95: Inaccurate encodings with Webpack.
I was having a weird issue with helium-js earlier this week. I was taking an account's bin and putting it to a b58 address. I realized that it didn't match my bin to b58 address using another library. Investigating further it seems that this is an issue caused by webpack and its handling of checksumBytes of bs58CheckEncode(). I encountered this issue while running a simple React app, using react-scripts.

vulet:
This resolved #95 for me.

### DIFF
--- a/packages/crypto/src/__tests__/Utils.spec.ts
+++ b/packages/crypto/src/__tests__/Utils.spec.ts
@@ -1,0 +1,30 @@
+import { utils, Address } from '..'
+import { bobB58, usersFixture } from '../../../../integration_tests/fixtures/users'
+import * as bs58 from 'bs58'
+
+describe('bs58checkEncode', () => {
+  it('should encode a publickey payload to b58 address', async () => {
+    const { bob } = await usersFixture()
+    const address = new Address(0, 1, bob.publicKey)
+  	const vPayload = Buffer.concat([Buffer.from([0]), address.bin])
+    const checksum = utils.sha256(Buffer.from(utils.sha256(vPayload)))
+  	const checksumBytes = Buffer.alloc(4, checksum, 'hex')
+  	const result = Buffer.concat([vPayload, checksumBytes])
+  	const encoded = bs58.encode(result)
+  	expect(encoded).toBe(bobB58)
+  })
+})
+
+describe('bs58ToBin', () => {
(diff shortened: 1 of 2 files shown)


allenan:
hey @vulet thanks for looking into this. Was this just an issue with an intermediary state when deriving the b58 address from the public key? Just curious that we have existing tests around `bs58CheckEncode` that were passing before and continue to pass.

allenan:
oh I see it's refactored to need fewer conversions to `'hex'`? I can still get your tests to pass with the old version by adding a `'hex'`, but I can see how it'd be better to have fewer places where that's needed. But I'm not sure exactly how that resolves the issue you mentioned in #95 where there was a webpack issue?

vulet:
@allenan Thank you for the review, and I am sorry, I don't think I
explained my issue very well. My issue is encountered using Webpack/React,
so I am going to detail what has been happening to me while using
Webpack/React, as this doesn't seem to affect other Nodejs builds for me.
Locally, what is happening for me: the bs58CheckEncode's checksum is
currently being passed as a hex string. This is then used in checkSumBytes,
or checksumVerifyBytes in bs58ToBin(). This leads me to a checksumBytes or
checksumVerifyBytes of efbfbd06 for Bob, versus what I actually need, which
is c506d245 and obtainable by staying in buffer. I've got this working
locally, but this might be due to how I have set things up locally. I would
appreciate more time before being considered for merge as I'll be out for
the next couple of days and this PR might be too premature to throw into
master. Thanks, Michael.

allenan:
@vulet no problem. Let's double check this when you get back. Just want to be extra cautious merging anything around the crypto libraries. Enjoy your travels!

vulet:
Hi @allenan, thanks for waiting. This works for me, let me know if there's
anything that I can do to get this merged.

allenan:
@vulet sorry for the delay. Hope you enjoyed your time in the mountains! I was able to reproduce this in a webpack environment, and verified that this fix works. Thanks for the fix 🙏

vulet:
excellent @allenan, thanks for the merge!

PR #100, opened 2020-10-29
Return all predicted oracle prices

tyler-whitman:
Updates Oracle.getPredictedPrice() to match the API response and return an array of predicted prices.

https://developer.helium.com/blockchain/api/oracle#get-predicted-hnt-oracle-prices


allenan (approved):
looks good

Issue #99, opened 2020-10-26
Expose raw json from API resources


allenan:
It would be useful to store and expose the raw json that `@helium/http` gets back from the API, as well as parsing it into typescript classes.

For example, we could do:

```ts
const client = new Client()
const list = client.hotspots.list
const [hotspot] = list.take(1)

console.log(hotspot.toJSON())
```

Currently the workaround is to parse/stringify the class, but then you need to `delete` the client reference out of it: https://github.com/helium/explorer/blob/master/jobs/sync_hotspots.js#L16-L17

allenan:
resolved by https://github.com/helium/helium-js/pull/105

PR #105, opened 2020-12-01
Create hotspot list json response.

matthewcarlreetz:
This PR adds `DataModel` which is an abstract class for all resources to extend from. These classes will be required to implement their own `data` function which returns their serializable data. When querying a resource list you can now optionally call `takeJSON` instead of `take`


packages/http/src/resources/Hotspots.ts
```diff
@@ -24,6 +25,20 @@ export default class Hotspots {
   }
 
   async list(params: ListParams = {}): Promise<ResourceList<Hotspot>> {
+    const { hotspots, cursor } = await this.fetchList(params)
+    const data = hotspots.map((d: HTTPHotspotObject) => new Hotspot(this.client, d))
+    return new ResourceList(data, this.list.bind(this), cursor)
+  }
+
+  async listData(params: ListParams = {}): Promise<ResourceList<HotspotData>> {
```
tyler-whitman, line 33: So what happens then if you call HotspotData.activity() or HotspotData.witnesses()? These use the client so they would fail right?

packages/http/src/models/DataModel.ts
```diff
@@ -0,0 +1,15 @@
...
+import type Client from '../Client'
+
+export default abstract class DataModel {
+  protected client: Client
+
+  constructor(client: Client) {
+    this.client = client
+  }
+
+  get data(): Omit<this, 'client'> {
```
allenan, line 11: this is cool, didn't know about `Omit`

packages/http/src/resources/Hotspots.ts
allenan: you'd be making an intentional choice to terminate your query/chain and just get the data. so you'd probably want to get back a type that reflects that choice maybe? something that doesn't indicate that you could call chaining methods on it? additionally, I could see this `listData` method being hoisted up to a parent class, like `data` in `DataModel`

matthewcarlreetz: I like the idea of moving the list functions into a base class. I'll investigate our options there.

packages/http/src/models/Transaction.ts
```diff
@@ -131,44 +224,46 @@ export default class Transaction {
         return this.toAddGatewayV1(json)
       case 'rewards_v1':
         return this.toRewardsV1(json)
+      default:
       case 'poc_receipts_v1':
         return this.toPocReceiptsV1(json)
-      default:
-        return prepareTxn(json)
+      // default:
```
tyler-whitman, line 230: Why did you comment this one out?

tyler-whitman:
Looks good to me, so this shouldn't change the existing calls right? It just adds the ability to do `.data()`

packages/http/src/models/Transaction.ts
matthewcarlreetz: I wanted to check with @allenan if it was ok to do this. Having that rogue object in there breaks my whole scheme of having the classes all extend from `DataModel`

matthewcarlreetz:
> Looks good to me, so this shouldn't change the existing calls right? It just adds the ability to do `.data()`

Yeah, it just adds a data function to all the model classes. When querying a list you can either `take(10)` or a `takeJson(10)`

packages/http/src/models/Transaction.ts
allenan: so this would assume that by default, any non-specified txn is a poc receipts txn? not sure that would work

tyler-whitman (approved):
Looks good

allenan (approved):
looks good to me

allenan:
resolves #99

PR #106, opened 2020-12-04
Add TransferHotspotV1 transaction

tyler-whitman:
- adds the new TransferHotspotV1 transaction
- updates @helium/proto to v1.2.0


allenan (approved):
looks good!

packages/transactions/src/TransferHotspotV1.ts
```diff
@@ -0,0 +1,126 @@
...
+  public amountToSeller?: number
+
+  public fee?: number
+
+  constructor(opts: TransferHotspotOptions) {
+    super()
+    this.gateway = opts.gateway
+    this.seller = opts.seller
+    this.buyer = opts.buyer
+    this.sellerSignature = opts.sellerSignature
```
allenan, line 49: I think the signatures need to come after the `calculateFee()` line because that will null them out for the fee calculation. Also, in PaymentV2, I added a condition where if the fee is passed in, it just uses that and skips the fee calculation

packages/transactions/src/__tests__/TransferHotspotV1.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  it('deserializes from a base64 string', async () => {
+    const transferHotspot = await transferHotspotFixture()
+    const transferHotspotString = transferHotspot.toString()
+    const deserialized = TransferHotspotV1.fromString(transferHotspotString)
+    expect(deserialized.gateway?.b58).toBe(transferHotspot.gateway?.b58)
+    expect(deserialized.buyer?.b58).toBe(transferHotspot.buyer?.b58)
+    expect(deserialized.seller?.b58).toBe(transferHotspot.seller?.b58)
+    expect(deserialized.amountToSeller).toBe(transferHotspot.amountToSeller)
+    expect(deserialized.buyerNonce).toBe(transferHotspot.buyerNonce)
+    expect(deserialized.fee).toBe(transferHotspot.fee)
```
allenan, line 62: we should probably add expects for the signatures here too

packages/transactions/src/TransferHotspotV1.ts
tyler-whitman: Makes sense, updating

packages/transactions/src/__tests__/TransferHotspotV1.spec.ts
tyler-whitman: Added

allenan (approved):
new changes look good

PR #107, opened 2020-12-04
Address#fromBin, PaymentV2#fromString

allenan:
- `@helium/crypto` adds `Address#fromBin` to construct and `Address` from its binary representation
- `@helium/transactions` adds `PaymentV2#fromString` to deserialize a PaymentV2 from its serialized string form


tyler-whitman (approved):
Looks good

PR #112, opened 2020-12-08
Fix 0 Amount Transfers

tyler-whitman:
When the amount is 0 and we pass it into the proto for serialization the default value is actually getting serialized into the proto (which is not expected) and on the blockchain it doesnt get serialized in for default values which is what is supposed to happen.

For this reason transfers with 0 amount were failing. When we dont include these 0 values proto uses the default 0 and doesnt include them in the serialized transaction which makes everything line up correctly.


allenan (approved):

lthiery:
This was pretty interesting to debug. I would think that the js proto serializer should omit fields when value is default but that doesn't seem the case. Unsure if its a misconfiguration, a bug in the js proto serializer, or whether its "just the way it is"

PR #113, opened 2020-12-09
Create Proto Ble Project

matthewcarlreetz:


packages/proto-ble/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "proto-ble",
+  "version": "3.9.1",
```
tyler-whitman, line 3: Should we start it at 1.0.0?

allenan: I think it'll get autolinked to the versions of the others once I do the first publish. So shouldn't really matter

tyler-whitman (approved):

packages/proto-ble/package.json
tyler-whitman: Ok np

packages/proto-ble/src/__tests__/gateway.test.ts
```diff
@@ -0,0 +1,11 @@
+import { add_gateway_v1 as AddGatewayV1, Iadd_gateway_v1 } from '../../build'
+
+const dict: Iadd_gateway_v1 = { payer: 'tacos mcgee' }
+describe('wifi_remove_v1', () => {
```
allenan, line 4: i think this `describe` got copied/pasted around

packages/proto-ble/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "proto-ble",
```
allenan, line 2: name should be `"@helium/proto-ble"`

```suggestion
  "name": "@helium/proto-ble",
```

allenan (changes requested):
overall looks great! just a couple minor things then this can be merged and published

allenan (approved):
lg

PR #116, opened 2020-12-11
Feature/add gateway deserialize

matthewcarlreetz:


packages/transactions/src/AddGatewayV1.ts
```diff
@@ -17,23 +19,40 @@ interface SignOptions {
...
     super()
+
     this.owner = opts.owner
     this.gateway = opts.gateway
     this.payer = opts.payer
     this.stakingFee = 0
     this.fee = 0
-    this.fee = this.calculateFee()
-    this.stakingFee = Transaction.stakingFeeTxnAddGatewayV1
+
```
allenan, line 45: I think we'll also want to deserialize the signatures and add those to the constructor if they're provided. that way, when we deserialize the partial txn that the hotspot creates, we keep the signature it added to it

allenan: one note is to assign them in the constructor after the `fee` calc because that calculation wipes out the signatures in order to perform its calculation

packages/transactions/package.json
```diff
@@ -32,8 +32,6 @@
     "long": "^4.0.0",
     "path": "^0.12.7"
   },
-  "devDependencies": {
-    "@helium/crypto": "^3.1.0"
```
tyler-whitman, line 36: Should we add this as a peer dependency so that users know its required?

tyler-whitman (approved):

packages/transactions/package.json
matthewcarlreetz: It's required in `"dependencies"`
https://github.com/helium/helium-js/blob/afceb7fe6f07e178874a93af5ddc3e4b9ebea862/packages/transactions/package.json#L30

I didn't think it was needed in both? @allenan ?

allenan: I think it's fine as is, in dependencies

allenan (approved):
looks good, will publish

PR #117, opened 2020-12-14
[@helium/http] add support for transfer hotspot v1

allenan:


tyler-whitman (approved):

packages/http/src/models/Transaction.ts
```diff
@@ -214,6 +216,37 @@ function prepareTxn(txn: any) {
...
+
+  seller!: string
+
+  height!: number
+
+  hash!: string
+
+  gateway!: string
+
+  buyer_nonce!: number
```
tyler-whitman, line 232: Did we want this to be camel case?

allenan: yes, thanks!

PR #119, opened 2020-12-19
hotspot rewards http endpoint

tyler-whitman:


allenan (approved):
looks good

packages/http/src/models/Hotspot.ts
```diff
@@ -111,6 +112,10 @@ export default class Hotspot extends DataModel {
     return new Hotspots(this.client, this)
   }
 
+  public get rewards(): HotspotRewards {
```
allenan, line 115: actually, don't we need this to take min/max time?

allenan: oh nvm I see you'd do a `rewards.getSum()`

PR #124, opened 2021-01-05
Add invalidReason to invalid witnesses

danielcolinjames:
This PR adds the new `invalid_reason` property to witnesses, so that it can be displayed on the frontend when using helium-js for anywhere witnesses show up.

I'm new to TypeScript (and Jest) so hopefully I didn't mess anything up; happy to make edits!

### DIFF
--- a/packages/http/src/models/Challenge.ts
+++ b/packages/http/src/models/Challenge.ts
@@ -152,8 +152,8 @@ const constructPath = (path: HTTPPathObject[]): Path[] => {
     }
     return {
       witnesses: pathObject.witnesses.map(
-        (witness) => {
-          let witnessObject = {
+        (witness) =>
+          ({
             timestamp: witness.timestamp,
             snr: witness.snr,
             signal: witness.signal,
@@ -161,14 +161,12 @@ const constructPath = (path: HTTPPathObject[]): Path[] => {
             owner: witness.owner,
             location: witness.location,
             isValid: isValidWitness(witness),
+            ...(witness.hasOwnProperty('invalid_reason') && { invalidReason: witness.invalid_reason }),
             gateway: witness.gateway,
             frequency: witness.frequency,
             datarate: witness.datarate,
             channel: witness.channel,
-          } as Witness
-          if (!isValidWitness(witness)) witnessObject = { ...witnessObject, invalidReason: witness.invalid_reason }
-          return witnessObject
-        },
+          } as Witness),
       ) as Witness[],
       receipt: hasReceipt
         ? ({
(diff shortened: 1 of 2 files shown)


packages/http/src/models/Challenge.ts
```diff
@@ -150,19 +152,23 @@ const constructPath = (path: HTTPPathObject[]): Path[] => {
...
+            packetHash: witness.packet_hash,
+            owner: witness.owner,
+            location: witness.location,
+            isValid: isValidWitness(witness),
+            gateway: witness.gateway,
+            frequency: witness.frequency,
+            datarate: witness.datarate,
+            channel: witness.channel,
+          } as Witness
+          if (!isValidWitness(witness)) witnessObject = { ...witnessObject, invalidReason: witness.invalid_reason }
```
allenan, line 169: I think you could do this without defining and mutating an intermediate object:

```js
{
  timestamp: witness.timestamp,
  ...(!isValidWitness(witness)) && { invalidReason: witness.invalid_reason },
}
```

Also, is the `isValidWitness` function call necessary? or could you just check to see if `invalid_reason` is present in the json?

packages/http/src/models/__tests__/Challenge.spec.ts
```diff
@@ -130,6 +134,18 @@ describe('Challenge Model', () => {
...
+  it('has invalid reason when witness is not valid', () => {
+    const challenge = new Challenge(challengeJson([
+      {
+        witnesses: [mockWitness(false)] as HTTPWitnessesObject[],
+        receipt: mockReceipt,
+        geocode: mockGeocode,
+        ...mockPathData,
+      } as HTTPPathObject,
+    ] as HTTPPathObject[]))
+    expect(challenge.path[0].witnesses[0].invalidReason).not.toBeUndefined();
```
allenan, line 146: might want to test that `invalidReason` equals `fake_invalid_reason`

packages/http/src/models/Challenge.ts
danielcolinjames: oh yeah that's way better, I tried to do something like that but couldn't remember the syntax for it, nice! and yeah will just check for witness.invalid_reason instead, good call 👍

allenan (changes requested):
nice work with ts and jest! just a couple small questions/suggestions

packages/http/src/models/__tests__/Challenge.spec.ts
danielcolinjames: good call!

danielcolinjames:
updated! thanks for the suggestions :)

allenan (approved):

PR #125, opened 2021-01-05
Add unknown txn type.

matthewcarlreetz:


packages/http/src/models/Transaction.ts
```diff
@@ -202,7 +217,7 @@ export type AnyTransaction =
   | AssertLocationV1
   | PocReceiptsV1
   | TransferHotspotV1
-  | GenericDataModel
```
tyler-whitman, line 205: Is the GenericDataModel still used?

tyler-whitman (approved):

packages/http/src/models/Transaction.ts
matthewcarlreetz: Yeah, it's a base type for other things

PR #130, opened 2021-01-20
Use Balance for Hotspot Rewards

tyler-whitman:
Hotspot rewards now use Balance<NetworkToken> instead of raw numbers.

### DIFF
--- a/packages/http/src/models/HotspotReward.ts
+++ b/packages/http/src/models/HotspotReward.ts
@@ -27,7 +27,7 @@ interface HTTPHotspotRewardsSum {
 export type HotspotRewardData = HotspotReward
 
 function toBalance(floatValue: number): Balance<NetworkTokens> {
-  return new Balance(floatValue * 100000000, CurrencyType.networkToken)
+  return Balance.fromFloat(floatValue, CurrencyType.networkToken)
 }
 
 export default class HotspotReward extends DataModel {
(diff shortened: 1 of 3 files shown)


packages/http/src/models/HotspotReward.ts
```diff
@@ -25,39 +26,40 @@ interface HTTPHotspotRewardsSum {
 
 export type HotspotRewardData = HotspotReward
 
+function toBalance(floatValue: number): Balance<NetworkTokens> {
+  return new Balance(floatValue * 100000000, CurrencyType.networkToken)
```
allenan, line 30: do we have this `100000000` number stored somewhere central? Or a `Balance.fromFloat` function? Ideally this kind of math wouldn't be done at this level

allenan: actually maybe it's because this is the first instance of the API returning float balances rather than integer balances. could you add a `Balance.fromFloat` static method? the CurrencyType stores a `public coefficient: BigNumber` which should specify the number of zeros to multiply by

tyler-whitman: Good point. It looks like we can do this instead.

`floatValue / CurrencyType.networkToken.coefficient.toNumber()`

What do you think?

tyler-whitman: Or we can use the BigNumber library directly instead of doing that .toNumber, but then we need to pull the library into the http library

tyler-whitman: Ok its updated now

allenan (approved):
nice!

PR #136, opened 2021-01-28
Fix balances not being created due to wrong casing.

matthewcarlreetz:


packages/http/src/models/Transaction.ts
```diff
@@ -259,20 +259,20 @@ function prepareTxn(txn: any) {
     txn.fee = new Balance(txn.fee, CurrencyType.dataCredit)
   }
 
-  if (txn.stakingFee && typeof txn.stakingFee === 'number') {
-    txn.stakingFee = new Balance(txn.stakingFee, CurrencyType.dataCredit)
+  if (txn.staking_fee && typeof txn.staking_fee === 'number') {
```
tyler-whitman, line 262: Ah whoops we should have noticed this in review. Can we add maybe a `AnyHTTPTransaction` type to txn? I guess its kinda hard to type it since it could be any of the HTTP response ones?

tyler-whitman (approved):

tyler-whitman:
Seems like the build is failing, not sure if its these changes or something wrong with the build?

allenan:
just merged master to fix the failing test

allenan (approved):

PR #140, opened 2021-02-03
Witness sum and hotspot reward bucket data

tyler-whitman:


packages/http/src/resources/HotspotRewards.ts
```diff
@@ -1,11 +1,17 @@
...
+  HotspotRewardSum,
+  HTTPHotspotReward,
+  HTTPHotspotRewardSum,
+} from '../models/HotspotReward'
 import ResourceList from '../ResourceList'
 
 interface ListRewardsParams {
-  minTime?: Date
-  maxTime?: Date
+  minTime?: Date | string // Date or natural format eg: '-30 day' '-1 week' '-5 hour'
```
allenan, line 11: it looks like there might be a way to define the natural language param as a template literal type: https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

tyler-whitman: Nice, added a `NaturalDate` type

allenan (approved):
looks great

PR #141, opened 2021-02-04
export WitnessSum in the http index

tyler-whitman:


tyler-whitman:
This is now irrelevant with the new sum class

PR #142, opened 2021-02-04
http account rewards

allenan:
generalizes `HotspotRewards` into a `Rewards` resource and allows it to be applied to either hotspots or accounts


tyler-whitman (approved):

packages/http/src/models/Sum.ts
```diff
@@ -0,0 +1,52 @@
...
+  sum: number
+  stddev: number
+  min: number
+  median: number
+  max: number
+  avg: number
+}
+
+function floatToBalance(floatValue: number): Balance<NetworkTokens> {
+  return Balance.fromFloat(floatValue, CurrencyType.networkToken)
```
tyler-whitman, line 18: Can you add a check here

```
if (!number) return undefined
```

I think challenge sums dont have a total so we should leave that out for those ones

tyler-whitman:
Looks great

packages/http/src/models/Sum.ts
allenan: I started doing that, but figured it would be out of the scope of this PR and could be handled in the challenge PR

tyler-whitman: Ok sounds good

PR #143, opened 2021-02-04
add challenge sum data

tyler-whitman:
Adds the new challenge sum endpoint data to @helium/http

The new api route is `hotspot/address/challenges/sum`

```
    const client = new Client()
    const list = await client.hotspot('address').challenges.sum.list({ minTime, maxTime, bucket })
    const challengeSums = await list.take()
```


packages/http/src/models/Sum.ts
```diff
@@ -21,30 +22,65 @@ function floatToBalance(floatValue: number): Balance<NetworkTokens> {
...
-  public max: Balance<NetworkTokens>
+  public min: number
 
-  public avg: Balance<NetworkTokens>
+  public balanceMin: Balance<NetworkTokens>
 
-  constructor(client: Client, rewards: HTTPSum) {
+  public median: number
+
+  public balanceMedian: Balance<NetworkTokens>
```
allenan, line 39: maybe rather than public members that get populated during the constructor, they could be getters that are only constructed if called?

```ts
get balanceTotal() {
  return floatToBalance(this.total)
}
```

allenan: by making it a `get` it's called the same way: `sum.balanceTotal` but it's just computed at the time it's called

packages/http/src/models/Sum.ts
```diff
@@ -21,30 +22,65 @@ function floatToBalance(floatValue: number): Balance<NetworkTokens> {
...
+
+  public balanceAvg: Balance<NetworkTokens>
+
+  public sum: number
+
+  public balanceSum: Balance<NetworkTokens>
+
+  public type: SumsType
+
+  constructor(client: Client, rewards: HTTPSum, type: SumsType) {
```
allenan, line 55: does `rewards` make sense here as a variable name?

packages/http/src/models/Sum.ts
tyler-whitman: How do those TS getters work? Because I think it defines it as a property on the object anyway? Or do you call them as a function?

tyler-whitman: Ah ok, let me update them

allenan: you don't have to call them as a function, they are called like properties, but I believe they're computed at the time they're called. So you wouldn't have to do `balanceTotal()` you could just call `.balanceTotal` but you wouldn't have to compute balances for things where it doesn't make sense, like witnesses

packages/http/src/models/Sum.ts
tyler-whitman: Good point, changed it to `data`

allenan (approved):

PR #146, opened 2021-02-21
currency locale formatting

tyler-whitman:


allenan (approved):
nice, been needing a showTicker = false option for awhile now :)

PR #156, opened 2021-04-08
Add hotspots.elected(block)

danielcolinjames:
Adds `client.hotspots.elected(block)` for hotspots elected at a given block height, and `client.hotspots.elected()` for currently elected hotspots

Seems to pass the tests:
```sh
  elected
    ✓ retrieves elected hotspots at a given block height (6 ms)
    ✓ retrieves currently elected hotspots (5 ms)
```

### DIFF
--- a/packages/http/src/resources/Hotspots.ts
+++ b/packages/http/src/resources/Hotspots.ts
@@ -60,4 +60,13 @@ export default class Hotspots {
     } = await this.client.get(url)
     return new Hotspot(this.client, hotspot)
   }
+  async elected(block: number): Promise<Hotspot[]> {
+    const url = `/elected/${block}`
+    const response = await this.client.get(url)
+    const {
+      data: { data: hotspots },
+    } = response
+    const data = hotspots.map((h: HTTPHotspotObject) => new Hotspot(this.client, h))
+    return data
+  }
 }
--- a/packages/http/src/resources/Hotspots.ts
+++ b/packages/http/src/resources/Hotspots.ts
@@ -60,8 +60,9 @@ export default class Hotspots {
     } = await this.client.get(url)
     return new Hotspot(this.client, hotspot)
   }
-  async elected(block: number): Promise<Hotspot[]> {
-    const url = `/elected/${block}`
+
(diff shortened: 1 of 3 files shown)


packages/http/src/resources/Hotspots.ts
```diff
@@ -60,4 +60,14 @@ export default class Hotspots {
     } = await this.client.get(url)
     return new Hotspot(this.client, hotspot)
   }
+
+  async elected(block?: number): Promise<Hotspot[]> {
```
tyler-whitman, line 64: Does this us a cursor? Usually we follow the ResourceList pattern for array to support cursors in the API and @allenan asked me to do that even for non cursor APIs and then we can just send a MAX value. 

Should we update this one to use ResourceList as well @allenan

tyler-whitman (changes requested):

packages/http/src/resources/Hotspots.ts
danielcolinjames: ah okay yeah was considering that but wasn't sure if it made sense here or not

allenan: yeah if we return a list of resources, we should use the `ResourceList` class so that the iteration methods exposed are consistent

packages/http/src/resources/__tests__/Hotspots.spec.ts
```diff
@@ -172,6 +174,67 @@ describe('get', () => {
...
+        hotspotFixture({ name: 'current-consensus-hotspot-14' }),
+        hotspotFixture({ name: 'current-consensus-hotspot-15' }),
+        hotspotFixture({ name: 'current-consensus-hotspot-16' }),
+      ],
+    })
+
+  it('retrieves elected hotspots at a given block height', async () => {
+    const client = new Client()
+    const elected = await client.hotspots.elected(123456)
+    expect(elected[0].name).toBe('previous-consensus-hotspot-1')
```
allenan, line 226: might also want to test that for example `.take` works after adding ResourceList above

allenan (approved):

PR #157, opened 2021-04-12
Allow disabling retries

danielcolinjames:
This adds an Options object to the constructor of Client, so that when initializing the Client, you could specify to not retry failed fetches. And obviously more options could be added in the future.

I'm not sure how to test out this implementation, as all the tests I wrote were failing

### DIFF
--- a/packages/http/src/Client.ts
+++ b/packages/http/src/Client.ts
@@ -34,22 +34,29 @@ interface BlockFromHeightOrHashFn {
   (heightOrHash: number | string): Block
 }
 
+interface Options {
+  retry: number | false
+}
+
 export default class Client {
   public network!: Network
 
   private axios!: AxiosInstance
 
-  constructor(network: Network = Network.production) {
+  constructor(network: Network = Network.production, options: Options = { retry: 5 }) {
     this.network = network
     this.axios = axios.create({
       baseURL: this.network.endpoint,
     })
-    this.axios.defaults.raxConfig = {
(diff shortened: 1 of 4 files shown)


packages/http/src/resources/Hotspots.ts
```diff
@@ -60,4 +60,14 @@ export default class Hotspots {
     } = await this.client.get(url)
     return new Hotspot(this.client, hotspot)
   }
+
+  async elected(block?: number): Promise<ResourceList<Hotspot>> {
```
allenan, line 64: why is this present in this PR?

packages/http/src/Client.ts
```diff
@@ -34,22 +34,29 @@ interface BlockFromHeightOrHashFn {
...
+  constructor(network: Network = Network.production, options: Options = { retry: 5 }) {
     this.network = network
     this.axios = axios.create({
       baseURL: this.network.endpoint,
     })
-    this.axios.defaults.raxConfig = {
-      instance: this.axios,
-      retry: 5,
-      noResponseRetries: 5,
+    let { retry } = options
```
allenan, line 51: I might make `retry` a public property of the class, like `network`, that way in the tests you can test that it gets configured correctly

packages/http/src/resources/Hotspots.ts
danielcolinjames: woops, forked from the last branch I was working on instead of from master

PR #158, opened 2021-04-13
Allow disabling retry

danielcolinjames:
Making a new PR instead of #157 because that one had commits from another branch in it by accident.

Still need to figure out how to properly test multiple `nock` requests as I don't think it's working properly here.

From some googling, it sounds like maybe `nock` and `axios` [sometimes don't play nicely together](https://github.com/nock/nock/issues/699#issuecomment-255860434), but I'm not sure if that's what's happening here or not. I also found [this library](https://github.com/axios/moxios) which might be useful if that is indeed what's happening.

But despite difficulties with testing, I'm pretty sure the logic works.

### DIFF
--- a/packages/http/src/Client.ts
+++ b/packages/http/src/Client.ts
@@ -34,22 +34,29 @@ interface BlockFromHeightOrHashFn {
   (heightOrHash: number | string): Block
 }
 
+interface Options {
+  retry: number | false
+}
+
 export default class Client {
   public network!: Network
 
   private axios!: AxiosInstance
 
-  constructor(network: Network = Network.production) {
+  constructor(network: Network = Network.production, options: Options = { retry: 5 }) {
     this.network = network
     this.axios = axios.create({
       baseURL: this.network.endpoint,
     })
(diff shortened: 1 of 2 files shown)


packages/http/src/Client.ts
```diff
@@ -34,22 +34,29 @@ interface BlockFromHeightOrHashFn {
...
 }
 
+interface Options {
+  retry: number | false
+}
+
 export default class Client {
   public network!: Network
-
+  public retry!: number | false
```
tyler-whitman, line 43: I'm wondering if instead of false, we just keep it a number and you would pass 0 for no retries?

tyler-whitman (changes requested):

packages/http/src/Client.ts
danielcolinjames: yeah that's probably cleaner, I actually saw axios or another library had that same pattern somewhere, good call 👍

allenan (approved):

tyler-whitman (approved):

PR #159, opened 2021-04-13
Add ability to search hotspots by name

matthewcarlreetz:


packages/http/src/resources/Hotspots.ts
```diff
@@ -34,19 +35,22 @@ export default class Hotspots {
     params: ListParams = {},
   ): Promise<{ hotspots: HTTPHotspotObject[]; cursor?: string }> {
     let url = '/hotspots'
+    const clientParams:{cursor?:string, search?:string} = { cursor: params.cursor }
```
allenan, line 38: hm why does `clientParams` have `search` as an optional attribute, but doesn't populate it?

matthewcarlreetz: If it's present, it gets populated on line 51

packages/http/src/resources/__tests__/Hotspots.spec.ts
```diff
@@ -283,6 +283,23 @@ describe('listJson', () => {
...
+  nock('https://api.helium.io')
+    .get('/v1/hotspots/name')
+    .query({ search: 'chicken-burrito' })
+    .reply(200, {
+      data: [hotspotFixture({ name: 'chicken-burrito-guacamole' }), hotspotFixture({ name: 'chicken-burrito-salsa' })],
+    })
+
+  it('lists hotspots', async () => {
+    const client = new Client()
+    const list = await client.hotspots.list({ search: 'chicken-burrito' })
```
allenan, line 296: i'm just wondering if we want to overload the `list` function here, or maybe create a separate `search` function so the syntax would be like

```ts
client.hotspots.search({ name: 'chicken-burrito' })
```

I'm also ok with it as it is, just maybe something to consider

matthewcarlreetz: I like that idea. If that feels better to you, I will update it.

allenan: yeah I guess it might also help simplify the growing list function, if we separated out search. it seems a little complex right now, and maybe odd to need to separate out the cursor from the other params

matthewcarlreetz: cool. I'll change it.

packages/http/src/resources/Hotspots.ts
```diff
@@ -24,6 +24,16 @@ export default class Hotspots {
     return new Hotspot(this.client, { address })
   }
 
+  async search(term:string): Promise<ResourceList<Hotspot>> {
```
matthewcarlreetz, line 27: Is this more like what you were thinking @allenan?

allenan: this looks good yeah. definitely keeps the branching logic in the `fetchList` function simpler 👍

allenan (approved):

PR #161, opened 2021-04-14
assert v2 support in http

tyler-whitman:
Adds assert v2 support to @helium/http. Once https://github.com/helium/blockchain-http/issues/208 is done there will be a couple more changes to add.

### DIFF
--- a/packages/http/src/resources/__tests__/Transactions.spec.ts
+++ b/packages/http/src/resources/__tests__/Transactions.spec.ts
@@ -263,13 +263,16 @@ describe('list from hotspot', () => {
     expect(txn4 instanceof AssertLocationV2).toBeTruthy()
 
     expect((txn0 as AssertLocationV1).hash).toBe('fake-hash-1')
+    expect((txn0 as AssertLocationV1).data.hash).toBe('fake-hash-1')
     expect((txn1 as AddGatewayV1).hash).toBe('fake-hash-2')
+    expect((txn1 as AddGatewayV1).data.hash).toBe('fake-hash-2')
     expect((txn1 as AddGatewayV1).stakingFee.toDataCredits().toString()).toBe('1 DC')
     expect((txn2 as UnknownTransaction).time).toBe(1587299256)
     expect((txn3 as TokenBurnV1).fee.toDataCredits().toString()).toBe('35,000 DC')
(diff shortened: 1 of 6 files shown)


packages/http/src/resources/__tests__/Transactions.spec.ts
```diff
@@ -226,27 +227,52 @@ describe('list from hotspot', () => {
...
+    expect((txn0 as AssertLocationV1).data.hash).toBe('fake-hash-1')
     expect((txn1 as AddGatewayV1).hash).toBe('fake-hash-2')
+    expect((txn1 as AddGatewayV1).data.hash).toBe('fake-hash-2')
     expect((txn1 as AddGatewayV1).stakingFee.toDataCredits().toString()).toBe('1 DC')
     expect((txn2 as UnknownTransaction).time).toBe(1587299256)
     expect((txn3 as TokenBurnV1).fee.toDataCredits().toString()).toBe('35,000 DC')
+    expect((txn4 as AssertLocationV2).hash).toBe('fake-hash-4')
+    expect((txn4 as AssertLocationV2).gain).toBe(12)
+    expect((txn4 as AssertLocationV2).elevation).toBe(0)
+    expect((txn4 as AssertLocationV2).data.elevation).toBe(0)
```
danielcolinjames, line 275: I'm not sure what the difference is between `.data.elevation` and `.elevation` or why they're both present here (but I'm not saying there's anything wrong, I'm sure there's a reason—I'm just curious why there's both)

danielcolinjames (approved):
looks good, passes tests, and the tests seem alright to me!

packages/http/src/resources/__tests__/Transactions.spec.ts
tyler-whitman: If I remember correctly we added the data stuff for storing in the app state. We were having issues with storing entire classes that had the client in them so I think we added this raw data accessor. 

I had to add those checks to meet the coverage requirements, but it should always be the same value as accessing normally.

danielcolinjames: ahh nice, I see, that makes sense, thanks for explaining it!

allenan (approved):
easy enough!

PR #162, opened 2021-04-15
Add counts to stats.

matthewcarlreetz:


tyler-whitman (approved):
looks good

packages/http/src/resources/__tests__/Stats.spec.ts
```diff
@@ -52,3 +61,16 @@ describe('get stats', () => {
...
+
+describe('get counts', () => {
+  nock('https://api.helium.io').get('/v1/stats/counts').reply(200, {
+    data: countsFixture(),
+  })
+
+  it('retrieves network count stats', async () => {
+    const client = new Client()
+    const counts = await client.stats.counts()
+    expect(counts.blocks).toBe(805339)
```
danielcolinjames, line 73: might want to check `.consensusGroups` too, just to double check that the camelCasing is working because it's the only two-word field

danielcolinjames (approved):
left one small suggestion for another thing you could add to the test but looks good 👍

packages/http/src/resources/__tests__/Stats.spec.ts
allenan: added the `expect`, good shout

allenan (approved):

PR #165, opened 2021-04-21
add witness info to witness model

allenan:
I was hoping to just have the `Witness` model extend the `Hotspot` model, but ran into some issues with dependency cycles. So just duplicated the `Hotspot` model and added the witness-specific fields


tyler-whitman (approved):
Yeah would have been nice to extend, but this works. Looks good

PR #167, opened 2021-05-03
payment v2 memos

tyler-whitman:
Adds support for memos in payment v2 transactions.

The memo is passed around in Helium JS and the API as a base64 encoded string, but is represented as a byte long on the blockchain.

The flow will be:
1. User enters up to 8 bytes of text in the payment memo field
2. We encode that text from utf8 to base64 and send it to helium-js. Helium-js encodes the base 64 to a long representation fo the bytes.
3. The API sends us the base64 string and we can decode it back to utf8 to get the plain text.

### DIFF
--- a/integration_tests/yarn.lock
+++ b/integration_tests/yarn.lock
@@ -2,47 +2,29 @@
 # yarn lockfile v1
 
 
-"@helium/crypto@^3.7.0":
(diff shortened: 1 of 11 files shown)


allenan (approved):
looks good to me

PR #168, opened 2021-05-05
Add listenAddrs to hotspot status.

matthewcarlreetz:


packages/http/src/models/Hotspot.ts
```diff
@@ -25,7 +25,12 @@ export interface HTTPHotspotObject {
   block_added?: number
   geocode?: HTTPGeocodeObject
   address: string
-  status?: Status
```
danielcolinjames, line 28: should there be a different snake-cased-fields interface called HTTPStatusObject to match the way HTTPGeocodeObject is done? not a huge deal either way, and just trying to understand the code better (still getting up to speed with TypeScript and helium-js, so this comment is more for my own understanding)

danielcolinjames (approved):
looks good!

packages/http/src/models/Hotspot.ts
```diff
@@ -115,7 +121,12 @@ export default class Hotspot extends DataModel {
     this.lng = hotspot.lng
     this.lat = hotspot.lat
     this.block = hotspot.block
-    this.status = hotspot.status
+    this.status = {
+      gps: hotspot.status?.gps || '',
+      height: hotspot.status?.height || 0,
```
tyler-whitman, line 126: Does the API return 0 or null? Might be good to tell the difference between one that is undefined and one that is actually 0?

packages/http/src/models/Hotspot.ts
tyler-whitman: Or should we just update the type that is already there for Status?

matthewcarlreetz: The Status that was there is the camelCased status. For the HTTP object we need snake case. We could create an interface `HTTPStatusObject`, but seeing as though it is only used in this context I didn't. If we ever wanted to reuse it, then we definitely would want to create a separate interface.

packages/http/src/models/Hotspot.ts
matthewcarlreetz: I'm not certain, but I think in this scenario null and 0 have the same meaning (sync hasn't started)

packages/http/src/models/Hotspot.ts
tyler-whitman: Ah its only for the HTTP type, ok this is fine. Should we update that status object in the hotspot to include the new camel case parameters? I think its missing `listenAddrs` ?

tyler-whitman: Wait nevermind its there lol

tyler-whitman (approved):

packages/http/src/models/Hotspot.ts
```diff
@@ -60,6 +65,7 @@ interface Status {
   gps: string
   height: number
   online: string
+  listenAddrs: string[]
```
matthewcarlreetz, line 68: @tyler-whitman you mean this? ^

packages/http/src/models/Hotspot.ts
tyler-whitman: Yeah I think it should be ok

packages/http/src/models/Hotspot.ts
tyler-whitman: Haha yeah, missed it sorry.

allenan (approved):

PR #187, opened 2021-06-16
Type Cleanup

tyler-whitman:
- update natural date type. I changed it to a string and left a comment as I don't think it was working correctly as it was. For some reason it complies fine in the library but wont compile in the hotspot app.
- update listen_addrs for the witness object to camel case


danielcolinjames (approved):
nice, looks good 👍

allenan (approved):
sad that we couldn't keep that more specific NaturalDate type, but maybe we can figure out a way to get the app to compile it. maybe need to update the app's TS version?

PR #193, opened 2021-07-09
feat(crypto): add support for bip39

### LINKED ISSUE #188: Mnemonic phrase should BIP39 compliant (generate a real checksum)
The current mobile wallet generates a 12-word mnemonic phrase that always has a `0000` checksum. This isn't BIP39 compliant. I noticed this while adding support for real BIP39 phrases into helium-wallet-rs. As the Helium ecosystem gets bigger, and we include things like ledger wallet I can see this being more and more of an issue. IMO, it is reasonable to expect a mnemonic phrase for a crypto wallet to be BIP39 compliant. 

Right now, the CLI wallet supports importing both mobile and CLI, and I have a PR for exporting a mnemonic phrase for an existing wallet. Both of these code paths, have to support "mobile" and "bip39" checksums in order to not surprise users.

Having to support two paths seems like a bad way forward. But, we have to be careful about the upgrade path. A suggestion from Discord:

* Fix this library (and then update the mobile wallet) so it always output only a BIP39-compliant  mnemonic phrase, rather than a phrase with a `0000` checksum.
* It has to be backwards compatible, so that users of "old mobile wallets" can still import their phrases. I don't think notifying users and expecting all will "update" their offline copy of their phrase is good enough. So, on import, the mobile wallet (using this library) should converting both mobile and BIP39 phrases into entropy. 

Then eventually maybe the CLI wallet can drop support for importing mobile phrases and only import BIP39 phrases. Or at the very least, the CLI wallet will not output mobile-type phrases on output (there's a PR on helium-wallet-rs to optionally output phrases in the verify subcommand)

One note, I'm not 100% sure, but AFAICT, a given secret can output both a mobile-type and BIP39 type phrase. Re-importing either into the mobile wallet will result in the same secret, which results in the same wallet and same public address. To put it another way, the mnemonic phrase checksum isn't part of the entropy stored that makes up the keypair. So changing just the checksum calculation and using that when showing the phrase to the user should not (again AFAICT) change the actual keypair that gets derived from that phrase on import. 

To make this more transparent to the mobile app user the `entropyFromMnemonic()` path  could just do (pseudocode)
```
If foundChecksum ==  '0000' or foundChecksum == bip39CalcChecksum:
   create entropy from this phrase.
```
This weakens the bip39 checksum verification a bit, but...maybe doesn't matter?

Note: I commented on #120 because I thought it was about this issue. But I don't think it is.

vulet:
This moves mnemonics forward to use the standardized bip39 checksum, and also allows for the legacy generated mnemonics. The [`crypto-react-native`](https://github.com/helium/helium-js/tree/master/packages/crypto-react-native) package hasn't changed, so this doesn't address #188, but perhaps can be used as reference for it.

### DIFF
--- a/integration_tests/fixtures/users.ts
+++ b/integration_tests/fixtures/users.ts
@@ -39,6 +39,8 @@ export const aliceWords = [
 export const bobB58 = '13M8dUbxymE3xtiAXszRkGMmezMhBS8Li7wEsMojLdb4Sdxc4wc'
 export const aliceB58 = '148d8KTRcKA5JKPekBcKFd4KfvprvFRpjGtivhtmRmnZ8MFYnP3'
 
+export const bobBip39Words = bobWords.map(word => word !== 'energy' ? word : 'episode')
+
 export const usersFixture = async () => ({
   bob: await Keypair.fromWords(bobWords),
   alice: await Keypair.fromWords(aliceWords),
--- a/integration_tests/tests/create_and_submit_payment.spec.ts
+++ b/integration_tests/tests/create_and_submit_payment.spec.ts
@@ -2,7 +2,7 @@ import nock from 'nock'
 import { Keypair, Address } from '@helium/crypto'
 import { Client } from '@helium/http'
 import { PaymentV1, PaymentV2 } from '@helium/transactions'
-import { bobWords, aliceB58 } from '../fixtures/users'
+import { bobWords, bobBip39Words, aliceB58 } from '../fixtures/users'
 
 test('create and submit a payment txn', async () => {
 
@@ -72,4 +72,28 @@ test('create and submit a PaymentV2 txn', async () => {
   const pendingTxn = await client.transactions.submit(serializedTxn)
 
   expect(pendingTxn.hash).toBe('txn hash')
+})
+
+test('using the bip39 checksum word should match serialization', async () => {
+
+  const bob = await Keypair.fromWords(bobBip39Words)
+  const aliceAddress = Address.fromB58(aliceB58)
+
+  const paymentTxn = new PaymentV2({
+    payer: bob.address,
+    payments: [
+      {
+        payee: aliceAddress,
+        amount: 10,
+      },
+    ],
+    nonce: 1,
+  })
+
+  const signedPaymentTxn = await paymentTxn.sign({ payer: bob })
+  const serializedTxn = signedPaymentTxn.toString()
+
+  expect(serializedTxn).toBe(
+    'wgGOAQohATUaccIv7+wiMZNq0oJrIX7OOdn3f8bEljmSYpnDhpKVEiUKIQGcZZ1yPMHoEKcuePfer0c2qH8Q74/PyAEAtTMn5+5JpBAKIAEqQK88GjmG9CrESHVdcL//ZfWD+KsBnbKmZqKlx8oD89FUms7OjZNcL5NiQ4o0jREg+ahkjc2jX4SgKBBniM+QoAA='
+  )
 })
--- a/packages/crypto/src/Keypair.ts
+++ b/packages/crypto/src/Keypair.ts
@@ -44,7 +44,7 @@ export default class Keypair {
   static async fromMnemonic(mnenomic: Mnemonic, netType?: NetType): Promise<Keypair> {
     await sodium.ready
     const entropy = mnenomic.toEntropy()
-    const seed = Buffer.concat([entropy, entropy])
+    const seed = entropy.length === 16 ? Buffer.concat([entropy, entropy]) : entropy
 
     return Keypair.fromEntropy(seed, netType)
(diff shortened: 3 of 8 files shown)


kellybyrd:
It looks like this change plus the existing code will:
* Accept 12 or 24 word phrases that have either '0000' or real checksums at the end and turn this into 128 or 256 bits of entropy.
* Given some existing entropy turn that into a real BIP39 phrase of the right length.

Right? If so, great! I should change my PR on helium-wallet-rs to not worry about outputting "mobile compatible" phrases from an existing wallet.

vulet:
This is correct @kellybyrd, one small change is that I would say that it handles the 128 or 256 bits of entropy rather than creating it. There shouldn't be any changes to the actual entropy generation. Before merge, [Line 47](https://github.com/helium/helium-js/pull/193/commits/6ee1a25904d8061fbbed0a948c656b377d5a897f) takes 16 bytes of entropy and doubles those same bytes to make 32 bytes. This might be done perhaps to prevent a `libsodium` throw on 16 bytes which is of 12 words.  The consequence of this means that we are not able to use 32 bytes or 24 words of entropy; because this too will also be doubled from 32 to 64. This will also cause a `libsodium` throw. But if we check for length before doubling then we can see if this is necessary or not.

I think it's safe to continue with your PR on two notes. This hasn't been merged, and also, I believe the `mobile` phrases are generated with [`crypto-react-native`](https://github.com/helium/helium-js/tree/master/packages/crypto-react-native) rather than the [`crypto`](https://github.com/helium/helium-js/tree/master/packages/crypto-react-native) package which is what this pull request aims to address. A final note of handling between `mobile` to `bip39` phrases, if an implementation between these two gives you 24 words from 12 words then it's probably due to the same entropy being doubled twice.

kellybyrd:
>
> I think it's safe to continue with your PR on two notes. This hasn't been merged, and also, I believe the `mobile` phrases are generated with [`crypto-react-native`](https://github.com/helium/helium-js/tree/master/packages/crypto-react-native) rather than the [`crypto`](https://github.com/helium/helium-js/tree/master/packages/crypto-react-native) package which is what this pull request aims to address. A final note of handling between `mobile` to `bip39` phrases, if an implementation between these two gives you 24 words from 12 words then it's probably due to the same entropy being doubled.

Thanks for the feedback. My "accept real BIP39 phrases" PR went into helium-wallet-rs in the last release or two. My [current pending one](https://github.com/helium/helium-wallet-rs/pull/150)  is just about getting phrases back out of a wallet. Since I've been working with it, the wallet has always used 256-bits of entropy to create a keypair. It employed this same doubling technique if needed. So in the PR I just linked, I have a hack which detects if the private key is just 128-bits doubled, and then generate a 12 word phrase from the first 128-bits. 

I think I'll remove the ability to select bip39|mobile phrase output in that PR, always outputting bip39. Then I'll do a new PR to have the seed input of helium-wallet-rs behave like this code does, automatically accepting either 0000 or a real checksum.

Do you have an estimate on when this code will be taken up by the mobile apps and released to end-users?

vulet:
> Thanks for the feedback. My "accept real BIP39 phrases" PR went into helium-wallet-rs in the last release or two. My [current pending one](https://github.com/helium/helium-wallet-rs/pull/150) is just about getting phrases back out of a wallet. Since I've been working with it, the wallet has always used 256-bits of entropy to create a keypair. It employed this same doubling technique if needed. So in the PR I just linked, I have a hack which detects if the private key is just 128-bits doubled, and then generate a 12 word phrase from the first 128-bits.

> I think I'll remove the ability to select bip39|mobile phrase output in that PR, always outputting bip39. Then I'll do a new PR to have the seed input of helium-wallet-rs behave like this code does, automatically accepting either 0000 or a real checksum.

I think both of these are great.

> Do you have an estimate on when this code will be taken up by the mobile apps and released to end-users? 

No idea, I think since it's a crypto related PR things it might move slower.


I edited my OP comment to un-link your issue, as I actually forgot about the React Crypto library which will probably need some changes as well. Thanks @kellybyrd.

kellybyrd:
We talked on discord, here was the result:

Ok, I'm convinced the right thing to do is leave the CLI wallet more clunky (requiring user specify bip39|mobile) for now until the mobile wallet outputs bip39 all the time.

I'll think about another PR to make the CLI wallet more friendly on input. It can just detect mobile or bip39 checksum automatically. IIRC it was kind of messy to do that the first time I tried, I'll give it another attempt

allenan:
@vulet can you add test cases for this one as well? We'll want to make sure to test this extensively in the app as well before merging

vulet:
definitely, @allenan!

allenan (approved):
changes look good, we will also propagate this to the react native package 🙏

allenan:
this has been published as `@helium/crypto@3.38.0`

PR #200, opened 2021-07-25
added locationHex to challenge data model

radrob1:
Since location_hex is the res8 hex returned by the ETL API, it should be added to the challenge and witness data models/ interfaces. This can then be used in Explorer to calculate res8 distances.


danielcolinjames (changes requested):
nice, thanks for doing this @radrob1! just reviewing your attached PR on the Explorer repo.

for this PR could you just remove the shrinkwrap JSON file? other than that it looks good, but I'm not really familiar with shrinkwrap and I think we can revisit that in the future for this repo, or if you want to put that in a different PR that would be great too

also would be great if you could add some lines to the relevant test files for this PR just to double check that everything here will work the way we expect for the given API responses

packages/http/yarn.lock
```diff
@@ -2,13 +2,6 @@
 # yarn lockfile v1
 
 
-"@helium/currency@^3.25.1":
```
allenan, line 5: also wondering why the lock file changed for this 🤔

radrob1: I didn't even notice this and not sure why it would have changed it. Must be my lack of js knowledge showing. Gonna clean these two things up though.

danielcolinjames (approved):
looks good @radrob1! thanks for making those changes 😄   I added a test case just to check for those location fields to make sure they were mapping the way we'd expect and all looks good, tests passed and everything 👍 

by the way if you had any feedback on getting helium-js running locally or anything that was confusing about running the tests etc., that would be useful to know so we can improve the documentation (which could definitely be better and more comprehensive)

PR #201, opened 2021-07-26
Add mode to witness model.

matthewcarlreetz:


danielcolinjames (approved):
nice, looks good to me. maybe can add a test just to be extra sure?

PR #227, opened 2021-08-18
[http] add hotspots location distance

### LINKED ISSUE #215: [http] add query for hotspots by distance from location
```
data/hotspots.js:
  52    if (!lat || !lon) return []
  53:   const hotspots = await fetchAll('/hotspots/location/distance', {
  54      lat,
```

allenan:
closes #215


danielcolinjames (approved):
🔥

PR #234, opened 2021-09-13
add support for bip39 to crypto-react-native

matthewcarlreetz:
This brings over the changes in @helium/crypto to @helium/crypto-react-native


allenan (approved):
looks good and tested to have backwards compatibility 👍

vulet:
hey @matthewcarlreetz @allenan, sorry for this late reply. i meant to comment earlier but forgot. this will still error out on the bip39 checksum as the checksumBits will still be 0000, right? i'm not able to verify because my computer isn't able to handle the React Native tool-chain.

PR #239, opened 2021-09-29
add api docs

tyler-whitman:
Add api docs using https://typedoc.org/
auto deploys to docs branch using https://github.com/JamesIves/github-pages-deploy-action and github actions

You can regenerate them by running `yarn docs`.


README.md
```diff
@@ -197,5 +197,5 @@ client.transactions.submit(signedPaymentTxn.toString())
 
 ```
 
-> :warning: Note that oracle prices change over time. It's possible for a transaction to fail if the oracle price changes in between the time the transaction is conrstructed and when it is absorbed by the consensus group. The API exposes what the next oracle price will be at `https://api.helium.io
+> ⚠️ Note that oracle prices change over time. It's possible for a transaction to fail if the oracle price changes in between the time the transaction is conrstructed and when it is absorbed by the consensus group. The API exposes what the next oracle price will be at `https://api.helium.io
```
matthewcarlreetz, line 200: Little typo here "conrstructed" should be "constructed".

tyler-whitman: fixed

allenan:
is there any way to have this build on demand, rather than committing to source?

allenan:
maybe some typedoc/pages github action integration?

tyler-whitman:
> is there any way to have this build on demand, rather than committing to source?

Yeah let me look into it. I was using it to host with GH pages easily but maybe there is a better way.

allenan (approved):

PR #240, opened 2021-10-04
[http] add name property to client

allenan:
adds a name property to the client which sets a `x-client-name` header on the request when set. This is an optional param for now. In the future we could make it required and have the client throw an error on initialization if its name isn't set, but that would require a change to the call signature and require more work than necessary right now.


tyler-whitman (approved):

danielcolinjames (approved):
nice! looks good and tests pass ✅

PR #254, opened 2021-11-03
fetch only specific chain vars

tyler-whitman:
Due to an upcoming change to chain vars which will add 10+mb to the data, we will now only fetch specific vars by default.

I left the 4 required for transaction config in by default so that the API remains non breaking, or you can specify any specific vars your want in the params and those will be returned. Along with this I added a `getAll` to return all vars and a `getTransactionVars` to specifically get the ones needed by transaction config.


allenan (approved):
looks good

PR #258, opened 2021-12-06
Add /roles support

danielcolinjames:
adds ability to get roles of recent activity by using /roles instead of /activity

used like:
- `client.account('address').roles.list()`
- `client.hotspot('address').roles.list()`
- `client.validator('address').roles.list()`

### DIFF
--- a/packages/http/src/models/Role.ts
+++ b/packages/http/src/models/Role.ts
@@ -1,0 +1,44 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable max-classes-per-file */
+import DataModel from './DataModel'
+
+export interface RoleJsonObject {
+  type: string
+  time: number
+  role: string
+  height: number
+  hash: string
+}
+
+export class RoleObject extends DataModel {
+  type!: string
+
+  time!: number
+
+  role!: string
+
+  height!: number
+
+  hash!: string
+
+  constructor(data: Role) {
+    super()
+    Object.assign(this, data)
+  }
+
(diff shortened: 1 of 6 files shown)


packages/http/src/models/Role.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface RoleJsonObject {
+  type: string
+  time: number
+  role: string
+  height: number
+  hash: string
+}
+
+export class RoleObject extends DataModel {
```
allenan, line 13: does this need to be called `RoleObject`? I'd expect `Role`

packages/http/src/models/Role.ts
```diff
@@ -0,0 +1,44 @@
...
+    super()
+    Object.assign(this, data)
+  }
+
+  get data(): Role {
+    return this
+  }
+}
+
+export type AnyRole = RoleObject
```
allenan, line 34: you probably don't need this since a Role doesn't have different shapes like txns do

packages/http/src/models/Role.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export default class Role {
+  public static fromJsonObject(json: RoleJsonObject): AnyRole {
```
allenan, line 37: not sure you need these either. Role can just be built from a constructor. txns are trickier because we're turning lots of different types of input into the object

allenan (approved):
looks good

PR #267, opened 2022-01-12
fix react native checksum generation

tyler-whitman:
@vulet @kellybyrd We brought the bip39 changes over from `@helium/crypto` to `@helium/crypto-react-native`, but I noticed the checksum was still always returning `0000`. This caused a problem for 24 word mnemonics as those checksums would also always be `00000000`, `10000000` or `00000001` and fail the checksum verification (as we had the pass for legacy 12 word `0000` checksums those still worked). The checksum should be a random set of bits correct? 

I think I found the problem, we use a different sha256 library on react native, and calling the function the way it was returned a string, with the change to using sha256.digest() we get the proper byte array and the checksum is randomized again.

Just wanted to tag you guys and see if this all makes sense, thanks!


kellybyrd:
There really should be some unit tests to verify this works as expected. The helium-wallet-rs code has several for 128-bit and 256-bit keys that do both 0000 checksum  and BIP39 checksum and compared those against phrases.

tyler-whitman:
> There really should be some unit tests to verify this works as expected. The helium-wallet-rs code has several for 128-bit and 256-bit keys that do both 0000 checksum and BIP39 checksum and compared those against phrases.

Yes, let me add some more tests as well. I think the React Native ones are missing.

kellybyrd:
> Yes, let me add some more tests as well. I think the React Native ones are missing.
IMO, using the same keys and phrases from the heium-wallet-rs project would be awesome, then we'd know the implementations produce the same result. But...hmnmm...IIRC those tests actually test a b58 version of a key against a phrase, so not sure that works for this code.

tyler-whitman:
> > Yes, let me add some more tests as well. I think the React Native ones are missing.
> > IMO, using the same keys and phrases from the heium-wallet-rs project would be awesome, then we'd know the implementations produce the same result. But...hmnmm...IIRC those tests actually test a b58 version of a key against a phrase, so not sure that works for this code.

So there are already tests in the repo for `@helium/crypto` but we dont have them for `@helium/crypto-react-native` which mostly uses the same code but has a couple of different dependencies and needs minor tweaks for that. I am copying those tests over and fixing them so the RN code is test as well as the plain JS code.

packages/crypto-react-native/src/utils.ts
```diff
@@ -29,9 +29,9 @@ export const binaryToByte = (bin: string) => parseInt(bin, 2)
 export const deriveChecksumBits = (entropyBuffer: Buffer | string) => {
   const ENT = entropyBuffer.length * 8
   const CS = ENT / 32
-  const hash = sha256(entropyBuffer)
+  const hash = sha256.digest(entropyBuffer)
```
tyler-whitman, line 32: This was the fix for the checksums, `sha256(entropyBuffer)` returned a hex string representation but a byte array was expected.

allenan (approved):
thank you for bringing these changes over and adding tests 🙏

packages/crypto-react-native/src/utils.ts
vulet: @tyler-whitman, seems good, thank you for investigating this. I had a feeling that there were still some things during the original @helium/crypto-react merge https://github.com/helium/helium-js/pull/234#issuecomment-955271811, but wasn't able to verify as I am without a computer. also, kudos on bringing the tests to the react package. have a good week!

PR #268, opened 2022-01-13
Create onboarding package with client

matthewcarlreetz:


packages/onboarding/src/Client.ts
```diff
@@ -0,0 +1,66 @@
...
+
+type Response<T>={
+  code: number
+  data: T | null
+  success: boolean
+  errorMessage?: string
+  errors?: Array<any>
+}
+
+export default class Client {
```
allenan, line 13: could having this top level class named `Client` get confusing with the `Client` exported out of `@helium/http`?

matthewcarlreetz: Likely. I'll rename.

packages/onboarding/src/Client.ts
```diff
@@ -0,0 +1,66 @@
...
+  constructor(baseURL: string = DEWI_ONBOARDING_API_BASE_URL) {
+    this.axios = axios.create({
+      baseURL,
+    })
+  }
+
+  async execute<T>(method:Method, path: string, params: Object = {}) {
+    const query = qs.stringify(params)
+    let url = path
+    if (method === 'GET' && query.length > 0) {
```
allenan, line 25: could `GET` specific logic be handled in the `get` function? then there wouldn't be a need to add case logic here

matthewcarlreetz: yeah, I like that.

packages/onboarding/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "onboarding",
```
allenan, line 2: think this should be `@helium/onboarding`

allenan (approved):
looks good!

PR #285, opened 2022-03-02
extract Address into a new @helium/address package

allenan:
this will allow address logic to be used regardless of which crypto environment you are in (ie nodejs vs react native)


packages/crypto-react-native/src/Keypair.ts
```diff
@@ -1,9 +1,11 @@
 import Sodium from 'react-native-sodium'
+import Address, { KeyTypes, NetTypes } from '@helium/address'
 import Mnemonic from './Mnemonic'
-import Address from './Address'
-import { KeyType, ED25519_KEY_TYPE } from './KeyType'
-import { NetType, MAINNET } from './NetType'
+// import { KeyType, ED25519_KEY_TYPE } from './KeyType'
```
tyler-whitman, line 4: Should we remove these?

tyler-whitman (approved):
Looks good, we will have to refactor all the apps when we update. Do we want to do with a breaking change on this one?

allenan:
> Looks good, we will have to refactor all the apps when we update. Do we want to do with a breaking change on this one?

Yeah I was going to make it a major version

PR #290, opened 2022-04-10
MultisigAddress & MultisigSignature Implementation

jcronyn:
- Create multisig public key from other keys
- New multisig signature from multiple public keys / signatures
- Util: Sort public keys
- Util: Sort signatures
- Binary helpers
- Additional helper functions as needed
- Test cases:
  - Create valid multisig address
  - Invalid keys to create multisig address
  - Create valid multisig signature
  - Invalid signatures for multisig signature
  - M of N signatures pass/fail
  - Create multisig payment transaction and pass validation

Pending:
- [ ] Testnet testing pending testnet availability

### DIFF
--- a/packages/address/src/MultisigAddress.ts
+++ b/packages/address/src/MultisigAddress.ts
@@ -1,0 +1,102 @@
+import {
+  bs58M,
+  bs58N,
+  bs58Version,
+  bs58MultisigPublicKey,
+  bs58NetType,
+  byteToNetType,
+  byteToKeyType,
+  sortAddresses,
+  bs58KeyType,
+} from './utils'
+import { sha256 } from 'multiformats/hashes/sha2'
+import { MULTISIG_KEY_TYPE } from './KeyTypes'
+import { NetType, MAINNET} from './NetTypes'
+import Address from './Address'
+
+export class MultisigAddress extends Address {
+  public M!: number
+
+  public N!: number
+
+  public constructor(version: number, netType: NetType, M: number, N: number, publicKey: Uint8Array) {
+    if (M > 256) {
+      throw new Error('required signers cannot exceed 256')
+    }
+    if (N > 256) {
+      throw new Error('total signers cannot exceed 256')
+    }
+    if (M > N) {
+      throw new Error('required signers cannot exceed total signers')
+    }
+    super(version, netType, MULTISIG_KEY_TYPE, publicKey);
+    this.M = M
+    this.N = N
+  }
+
+  get bin(): Buffer {
+    return Buffer.concat([
+      // eslint-disable-next-line no-bitwise
+      Buffer.from([this.netType | this.keyType]),
+      Buffer.from(new Uint8Array([this.M])),
+      Buffer.from(new Uint8Array([this.N])),
+      Buffer.from(this.publicKey),
+    ])
+  }
+
+  static fromB58(b58: string): MultisigAddress {
+    const keyType = bs58KeyType(b58)
+    if (keyType !== MULTISIG_KEY_TYPE) {
+      throw new Error('invalid keytype for multisig address')
+    }
+    const version = bs58Version(b58)
+    const netType = bs58NetType(b58)
+    const M = bs58M(b58)
+    const N = bs58N(b58)
+    const publicKey = bs58MultisigPublicKey(b58)
+    return new MultisigAddress(version, netType, M, N, publicKey)
+  }
+
+  static fromBin(bin: Buffer): MultisigAddress {
+    const version = 0
+    const byte = bin[0]
+    const netType = byteToNetType(byte)
+    const keyType = byteToKeyType(byte)
+    if (keyType !== MULTISIG_KEY_TYPE) {
+      throw new Error('invalid keytype for multisig address')
+    }
+    const M = bin[1]
+    const N = bin[2]
+    const publicKey = bin.slice(3, bin.length)
+    return new MultisigAddress(version, netType, M, N, publicKey)
+  }
+
+  public static async create(addresses: Address[], M: number, netType?: NetType): Promise<MultisigAddress> {
+    const version = 0
+    if (!netType) {
+      netType = MAINNET
+    }
+
+    let multisigPubKeysBin = new Uint8Array()
+    for (const address of sortAddresses(addresses)) {
+      if (address.keyType === MULTISIG_KEY_TYPE) {
+        return Promise.reject(new Error('cannot craeate multisig with invalid child keytype'))
+      }
+      multisigPubKeysBin = new Uint8Array([...multisigPubKeysBin, ...address.bin])
+    }
+
+    const publicKey = (await sha256.digest(multisigPubKeysBin))
+    return new MultisigAddress(version, netType, M, addresses.length, publicKey.bytes)
+  }
+
+  static isValid(b58: string): boolean {
+    try {
+      MultisigAddress.fromB58(b58)
+      return true
+    } catch (error) {
+      return false
+    }
+  }
+}
+
+export default MultisigAddress
--- a/packages/address/src/__tests__/MultisigAddress.spec.ts
+++ b/packages/address/src/__tests__/MultisigAddress.spec.ts
@@ -1,0 +1,85 @@
+import Address from '..'
+import { MultisigAddress } from '../MultisigAddress'
+import {  
+  bobB58,
+  aliceB58,
+  bobAliceMultisig1of2B58,
+  bobAliceMultisig2of2B58,
+  testnetBobAliceMultisig2of2B58,
+} from '../../../../integration_tests/fixtures/users'
+import { TESTNET } from '../NetTypes'
+
+
+describe('multisig b58', () => {
+  it('returns a b58 check encoded representation of a multisig address', async () => {
+    const addressMultisig2of2 = await MultisigAddress.create([Address.fromB58(bobB58), Address.fromB58(aliceB58)], 2)
+    expect(addressMultisig2of2.b58).toBe(bobAliceMultisig2of2B58)
+  })
+
+  it('supports multisig addresses', () => {
+    const addressMultisig2of2 = MultisigAddress.fromB58(bobAliceMultisig2of2B58)
+    expect(addressMultisig2of2.b58).toBe(bobAliceMultisig2of2B58)
+  })
+})
+
+describe('bin', () => {
+  it('returns a binary representation of the multisig ddress', async () => {
+    const addressMultisig2of2 = await MultisigAddress.create([Address.fromB58(bobB58), Address.fromB58(aliceB58)], 2)
+    expect(addressMultisig2of2.bin[0]).toBe(2)
+  })
+})
+
+describe('fromBin', () => {
+  it('builds a MultisigAddress from a binary representation', async () => {    
+    const multisigAddress = await MultisigAddress.create([Address.fromB58(bobB58), Address.fromB58(aliceB58)], 2)
+    const multisigAddressFromBin = MultisigAddress.fromBin(multisigAddress.bin)
+    expect(multisigAddressFromBin.b58).toBe(multisigAddress.b58)
+  })
+})
+
+describe('fromB58', () => {
+  it('builds an Address from a b58 string', () => {
+    const multisigAddressFromB58 = Address.fromB58(bobAliceMultisig2of2B58)
+    expect(multisigAddressFromB58.b58).toBe(bobAliceMultisig2of2B58)
+  })
+})
+
+describe('unsupported child key types', () => {
+  it('throws an error if creating address with multisig key type', async () => {
+    expect(async () => {
+      await MultisigAddress.create([MultisigAddress.fromB58(bobAliceMultisig2of2B58), Address.fromB58(aliceB58)], 2)
+    }).rejects.toThrow()
+  })
+})
+
+describe('isValid', () => {
+  it('returns true if the address is valid and supported', () => {
+    expect(MultisigAddress.isValid(bobAliceMultisig2of2B58)).toBeTruthy()
+    expect(MultisigAddress.isValid(bobAliceMultisig1of2B58)).toBeTruthy()
+  })
+})
+
+describe('testnet addresses', () => {
+  it('decodes testnet addresses from b58', async () => {
+    const address = MultisigAddress.fromB58(testnetBobAliceMultisig2of2B58)
+    expect(address.netType).toBe(TESTNET)
+  })
+})
+
+describe('testnet addresses', () => {
+  it('decodes testnet addresses from b58', async () => {
+    const address = MultisigAddress.fromB58(testnetBobAliceMultisig2of2B58)
+    expect(address.netType).toBe(TESTNET)
+  })
+})
+
+describe('erlang interop', () => {
+  it('makes the same multisig key as the erlang lib ', async () => {
+    const keys = [
+      Address.fromB58('11MJXxoWFp2bMsqKM6QZin6ync9DQ3fjjFjUrFiRCaKunmBEBhK'),
+      Address.fromB58('11x7jP9yAnyk5jeYywmsYDFdYq5xvKLKjP2zjhGzCwDSQtxcUDt'),
+    ]
+    const address = await MultisigAddress.create(keys, 1)
+    expect(address.b58).toBe('1SVRdbaAev7zSpUsMjvQrbRBGFHLXEa63SGntYCqChC4CTpqwftTPGbZ')
+  }) 
+})
--- a/packages/address/src/__tests__/Utils.spec.ts
+++ b/packages/address/src/__tests__/Utils.spec.ts
@@ -22,7 +22,7 @@ describe('bs58ToBin', () => {
     const address = new Address(0, NetTypes.MAINNET, 1, bob.publicKey).b58
     const bin = bs58.decode(address)
     const vPayload = bin.slice(0, -4)
-    const checksum = bin.slice(-4)
+    const checksum = Buffer.from(bin.slice(-4))
     const checksumVerify = sha256(Buffer.from(sha256.digest(vPayload)))
     const checksumVerifyBytes = Buffer.alloc(4, checksumVerify, 'hex')
     expect(checksumVerifyBytes).toStrictEqual(checksum)
--- a/packages/address/src/index.ts
+++ b/packages/address/src/index.ts
@@ -5,6 +5,7 @@
  */
 
 export { default } from './Address'
+export { MultisigAddress } from './MultisigAddress'
 export * as NetTypes from './NetTypes'
 export * as KeyTypes from './KeyTypes'
 export * as utils from './utils'
--- a/packages/address/src/utils.ts
+++ b/packages/address/src/utils.ts
@@ -3,6 +3,7 @@ import { sha256 } from 'js-sha256'
 import bs58 from 'bs58'
 import { KeyType } from './KeyTypes'
 import { NetType } from './NetTypes'
+import Address from './Address'
 
 export const bs58CheckEncode = (version: number, binary: Buffer | Uint8Array): string => {
   const vPayload = Buffer.concat([
@@ -62,4 +63,33 @@ export const bs58PublicKey = (bs58Address: string): Buffer => {
   const bin = bs58ToBin(bs58Address)
   const publicKey = Buffer.from(bin).slice(1)
   return publicKey
+}
+
+export const bs58M = (bs58Address: string): number => {
+  const bin = bs58ToBin(bs58Address)
+  const M = bin[1]
+  return M
+}
+
+export const bs58N = (bs58Address: string): number => {
+  const bin = bs58ToBin(bs58Address)
+  const N = bin[2]
+  return N
+}
+
+export const bs58MultisigPublicKey = (bs58Address: string): Buffer => {
+  const bin = bs58ToBin(bs58Address)
+  const publicKey = Buffer.from(bin).slice(3)
+  return publicKey
+}
+
+export const sortAddresses = (addresses: Address[]): Address[] => {
+  const addressMap = addresses.map(address => {
+    const charCodeArray = Array.from(address.b58).map((character):number => {
+      return character.charCodeAt(0)
+    })
+    return { address: address, buffer: new Uint8Array(charCodeArray)}
+  })
+
+  return addressMap.sort((a, b) => Buffer.compare(a.buffer, b.buffer)).map(obj => obj.address)
 }
--- a/packages/crypto/src/MultisigSignature.ts
+++ b/packages/crypto/src/MultisigSignature.ts
@@ -1,0 +1,121 @@
+import Address, { KeyTypes, MultisigAddress } from '@helium/address'
+import { sortAddresses } from '@helium/address/build/utils';
+import { verifySignature } from './utils'
+const {MULTISIG_KEY_TYPE} = KeyTypes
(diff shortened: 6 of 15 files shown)


packages/transactions/src/PaymentV2.ts
```diff
@@ -51,10 +51,9 @@ export default class PaymentV2 extends Transaction {
     this.signature = opts.signature
   }
 
-  serialize(): Uint8Array {
+  serialize(forSigning: boolean = false): Uint8Array {
```
jcronyn, line 54: Probably need to update the other serialize methods to make multisig signing easier, if this pattern is desirable

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+  bs58N,
+  bs58Version,
+  bs58MultisigPublicKey,
+  bs58NetType,
+  byteToNetType,
+  byteToKeyType,
+  sortAddresses,
+  bs58KeyType,
+} from './utils'
+import { sha256 } from 'multiformats/hashes/sha2'
```
tyler-whitman, line 12: You have some lint errors in this file that can be cleaned up.

```ESLint: `multiformats/hashes/sha2` import should occur before import of `./utils`(import/order)```

File: packages/address/src/MultisigAddress.ts [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+import { MULTISIG_KEY_TYPE } from './KeyTypes'
+import { NetType, MAINNET} from './NetTypes'
+import Address from './Address'
+
+export class MultisigAddress extends Address {
+  public M!: number
+
+  public N!: number
+
+  public constructor(version: number, netType: NetType, M: number, N: number, publicKey: Uint8Array) {
```
tyler-whitman, line 22: ```ESLint: This line has a length of 102. Maximum allowed is 100.(max-len)```

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (M > 256) {
+      throw new Error('required signers cannot exceed 256')
+    }
+    if (N > 256) {
+      throw new Error('total signers cannot exceed 256')
+    }
+    if (M > N) {
+      throw new Error('required signers cannot exceed total signers')
+    }
+    super(version, netType, MULTISIG_KEY_TYPE, publicKey);
```
tyler-whitman, line 32: Can remove the semi colon

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+  sortAddresses,
+  bs58KeyType,
+} from './utils'
+import { sha256 } from 'multiformats/hashes/sha2'
+import { MULTISIG_KEY_TYPE } from './KeyTypes'
+import { NetType, MAINNET} from './NetTypes'
+import Address from './Address'
+
+export class MultisigAddress extends Address {
+  public M!: number
```
tyler-whitman, line 18: Are `M` and `N` standard terms? I am not familiar with multisig implementations, but if possible can we provide a more descriptive name for the variables?

tyler-whitman: If I understand this properly, it's also always two signatures? Is there possibility for more than two?

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (keyType !== MULTISIG_KEY_TYPE) {
+      throw new Error('invalid keytype for multisig address')
+    }
+    const M = bin[1]
+    const N = bin[2]
+    const publicKey = bin.slice(3, bin.length)
+    return new MultisigAddress(version, netType, M, N, publicKey)
+  }
+
+  public static async create(addresses: Address[], M: number, netType?: NetType): Promise<MultisigAddress> {
```
tyler-whitman, line 74: ```ESLint: This line has a length of 108. Maximum allowed is 100.(max-len)```

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+    const version = 0
+    if (!netType) {
+      netType = MAINNET
```
tyler-whitman, line 77: ```ESLint: Assignment to function parameter 'netType'.(no-param-reassign)``` 

You could set up a new variable within the function or pass `netType || MAINNET` on line 89

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    let multisigPubKeysBin = new Uint8Array()
+    for (const address of sortAddresses(addresses)) {
+      if (address.keyType === MULTISIG_KEY_TYPE) {
+        return Promise.reject(new Error('cannot craeate multisig with invalid child keytype'))
```
tyler-whitman, line 83: `craeate` should be `create`

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+    try {
+      MultisigAddress.fromB58(b58)
+      return true
+    } catch (error) {
+      return false
+    }
+  }
+}
+
+export default MultisigAddress
```
tyler-whitman, line 102: You can move this to the top just make line 17 `export default class MultisigAddress extends Address {`

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+  bs58MultisigPublicKey,
+  bs58NetType,
+  byteToNetType,
+  byteToKeyType,
+  sortAddresses,
+  bs58KeyType,
+} from './utils'
+import { sha256 } from 'multiformats/hashes/sha2'
+import { MULTISIG_KEY_TYPE } from './KeyTypes'
+import { NetType, MAINNET} from './NetTypes'
```
tyler-whitman, line 14: `ESLint: A space is required before '}'.(object-curly-spacing)`

packages/address/src/MultisigAddress.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  public static async create(addresses: Address[], M: number, netType?: NetType): Promise<MultisigAddress> {
+    const version = 0
+    if (!netType) {
+      netType = MAINNET
+    }
+
+    let multisigPubKeysBin = new Uint8Array()
+    for (const address of sortAddresses(addresses)) {
```
tyler-whitman, line 81: ```ESLint: iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations.(no-restricted-syntax)```

packages/address/src/__tests__/MultisigAddress.spec.ts
```diff
@@ -0,0 +1,85 @@
+import Address from '..'
+import { MultisigAddress } from '../MultisigAddress'
```
tyler-whitman, line 2: This can just be `import MultisigAddress from '../MultisigAddress'` with the default export change

packages/address/src/__tests__/MultisigAddress.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+import {  
+  bobB58,
```
tyler-whitman, line 4: There are a bunch of lint errors to fix here too. You can ignore any of the dependency cycle ones, we need a refactor to fix those but all others can be fixed.

packages/address/src/__tests__/Utils.spec.ts
```diff
@@ -22,7 +22,7 @@ describe('bs58ToBin', () => {
     const address = new Address(0, NetTypes.MAINNET, 1, bob.publicKey).b58
     const bin = bs58.decode(address)
     const vPayload = bin.slice(0, -4)
-    const checksum = bin.slice(-4)
+    const checksum = Buffer.from(bin.slice(-4))
```
tyler-whitman, line 25: Why was this change needed?

packages/address/src/index.ts
```diff
@@ -5,6 +5,7 @@
  */
 
 export { default } from './Address'
+export { MultisigAddress } from './MultisigAddress'
```
tyler-whitman, line 8: Change this import with the default export change

packages/address/src/utils.ts
```diff
@@ -63,3 +64,32 @@ export const bs58PublicKey = (bs58Address: string): Buffer => {
...
+}
+
+export const bs58MultisigPublicKey = (bs58Address: string): Buffer => {
+  const bin = bs58ToBin(bs58Address)
+  const publicKey = Buffer.from(bin).slice(3)
+  return publicKey
+}
+
+export const sortAddresses = (addresses: Address[]): Address[] => {
+  const addressMap = addresses.map(address => {
```
tyler-whitman, line 87: should be `(address)`

packages/address/src/utils.ts
```diff
@@ -63,3 +64,32 @@ export const bs58PublicKey = (bs58Address: string): Buffer => {
...
+    const charCodeArray = Array.from(address.b58).map((character):number => {
+      return character.charCodeAt(0)
+    })
+    return { address: address, buffer: new Uint8Array(charCodeArray)}
```
tyler-whitman, line 91: `address: address` can just be `address`

tyler-whitman: Also needs a space at the end before the `}`

packages/crypto/src/MultisigSignature.ts
```diff
@@ -0,0 +1,121 @@
+import Address, { KeyTypes, MultisigAddress } from '@helium/address'
```
tyler-whitman, line 1: Lots of lint errors to fix in this file

packages/crypto/src/MultisigSignature.ts
```diff
@@ -0,0 +1,121 @@
...
+import { sortAddresses } from '@helium/address/build/utils';
```
tyler-whitman, line 2: You should export this from `/packages/address/src/index.ts` so that it isnt imported from a build directory. It should be added to the import above after it is exported properly.

packages/crypto/src/MultisigSignature.ts
tyler-whitman, line 1: These changes will need to be brought over to the `crypto-react-native` package if we want to support this in the helium app. I am ok with doing that as a separate PR if wanted though.

packages/crypto/src/__tests__/MultisigSignature.spec.ts
```diff
@@ -0,0 +1,145 @@
+import { MultisigAddress } from '@helium/address'
+import { MultisigSignature } from '..'
+import { usersFixture } from '../../../../integration_tests/fixtures/users'
+
+describe('create', () => {
+    it('invalid if not enough signatures', async () => {
```
tyler-whitman, line 6: Bunch of lint errors to fix here too.

packages/crypto/src/utils.ts
```diff
@@ -28,3 +28,7 @@ export const deriveChecksumBits = (entropyBuffer: Buffer | string) => {
 
   return bytesToBinary([].slice.call(hash)).slice(0, CS)
 }
+
+export const verifySignature = (signature: Uint8Array, message:  Uint8Array, pubKey: Uint8Array) => {
```
tyler-whitman, line 32: `ESLint: This line has a length of 101. Maximum allowed is 100.(max-len)`

packages/transactions/src/PaymentV2.ts
tyler-whitman: I think instead of passing a boolean lets pass an object with optional params:

You can make this line:
` serialize({ includeSignature = false }): Uint8Array {`

And then in `/transactions/src/Transaction.ts` make line 16:
` abstract serialize({ includeSignature } : { includeSignature?: boolean }): Uint8Array`

packages/transactions/src/PaymentV2.ts
```diff
@@ -51,10 +51,9 @@ export default class PaymentV2 extends Transaction {
...
     const BlockchainTxn = proto.helium.blockchain_txn
-
-    const paymentV2 = this.toProto()
+    const paymentV2 = this.toProto(forSigning=forSigning)
```
tyler-whitman, line 56: This should just be `const paymentV2 = this.toProto(forSigning)` or `const paymentV2 = this.toProto(includeSignature)` if we change it

tyler-whitman (changes requested):
This is looking great, have some review feedback and mostly lint errors to resolve.

packages/address/src/MultisigAddress.ts
jcronyn: > Are `M` and `N` standard terms? I am not familiar with multisig implementations, but if possible can we provide a more descriptive name for the variables?

The `M` and `N` are taken from the reference implementation, but I agree they're nondescript and can be named better - will update.

> If I understand this properly, it's also always two signatures? Is there possibility for more than two?

I'm no expert but my understanding is that there is no theoretical limit to the number of signers (other than physical payload limitations on the blockchain side that would prevent a large number of accounts/signers).

packages/transactions/src/PaymentV2.ts
jcronyn: I swore that I fixed this 😆  a consequence of primarily developing in Python

packages/address/src/MultisigAddress.ts
xandkar: Yes, "m-of-n" is a de facto standard term in multisig land. Example: https://en.bitcoin.it/wiki/Multi-signature

xandkar: Yes, there's no theoretical limit on N and M is constrained to 0 < M <= N.

packages/transactions/src/PaymentV2.ts
jcronyn: @tyler-whitman - I can do that but the provided code isn't an optional param. When I implement as suggested, I have to pass an empty "options object" to the method. (`{}`). Just wanted to ask for you to clarify so we can align on the preferred method. Thanks!

packages/address/src/MultisigAddress.ts
jcronyn: @tyler-whitman - What's the accepted way to run lint on this project? I run `yarn lint` but am getting way more errors than I would expect...

packages/transactions/src/PaymentV2.ts
tyler-whitman: Yeah thats a good point, and we cant assign default values in an abstract function. I guess we would have to do something like:

`abstract serialize(includeSignature?: boolean): Uint8Array`

packages/crypto/src/__tests__/MultisigSignature.spec.ts
jcronyn: @tyler-whitman - What's the accepted way to run lint on this project? I run yarn lint but am getting way more errors than I would expect...

packages/transactions/src/PaymentV2.ts
jcronyn: I ended up completely removing the parameter. Not sure why I had it present in the first place, to be honest 🤔

packages/crypto/src/__tests__/MultisigSignature.spec.ts
tyler-whitman: Yeah there are a lot of older unresolved lint errors. Which IDE do you use? You should be able to integrate for specific files or you can run yarn lint on a specific file by changing the command in `./package.json` line 8. For example:

`"lint": "eslint --ext .ts ./packages/crypto/src/MultisigSignature.ts"`

tyler-whitman: You can also auto fix a lot of them by adding `--fix ` to the command

packages/address/src/__tests__/Utils.spec.ts
syuan100: This was from me. Tests were failing without this, not sure if this was environment specific though.

jcronyn: I'll see if I can remove it

jcronyn: @tyler-whitman - Looks required to me, this test fails without the update.

jcronyn:
> This is looking great, have some review feedback and mostly lint errors to resolve.

All comments and lint issues addressed @tyler-whitman - Should be good to go :) Please let me know if there's anything else that needs updating.

abhay:
> EDIT: Cannot test end-to-end in testnet as multisig is not enabled...

Multisig should work in testnet. what issues are you seeing?

syuan100:
@abhay miscommunication here. It works :) Just sent a payment to a multisig address here: https://testnet-explorer.helium.com/txns/hNMFq1t7CwwhRXOgL4uKb73i-_P8Oo0eVVtWXLhXYBU

tyler-whitman:
@jcronyn I made a PR to fix the lint errors, with this running `yarn lint` will give no errors. Can you please take a look and merge that into this.

https://github.com/jcronyn/helium-js/pull/2

tyler-whitman (approved):
Looks good to me, unit tests are passing and tests have been run on testnet. The last thing to do here would be to bring the changes over to `crypto-react-native` if we want MultiSig on mobile but for now I think it should be good to merge this and do that as another improvement. 

Along with this I also pushed a fix to all the lint errors so now we can check lint more easily.

@allenan and @matthewcarlreetz mind taking a look before we merge this.

jcronyn:
@tyler-whitman - Added a multi-signature example to the `README.md`. Feel free to review and let me know if you have any feedback. Thanks!

README.md
```diff
@@ -201,3 +201,58 @@ client.transactions.submit(signedPaymentTxn.toString())
...
+
+// get the speculative nonce for the keypair
+const account = await client.accounts.get(multisigAddress.b58)
+
+// construct a PaymentV2 txn to sign
+const paymentTxn = new PaymentV2({
+  payer: multisigAddress,
+  payments: [
+    {
+      payee: alice,
```
abhay, line 237: I think you want the destination address to be another party here.

jcronyn: It doesn't necessarily need to be, it's perfectly valid to send HNT from a multisig address to one of the wallets that are a member of that multisig adress.

syuan100: I think Abhay is just saying for clarity's sake so people don't get confused :) we should just use a random address

README.md
```diff
@@ -201,3 +201,61 @@ client.transactions.submit(signedPaymentTxn.toString())
...
+// Create signatures payload, a map of address to signature
+const signatures = new Map([[bob.address, await bob.sign(paymentTxn.serialize())]])
+
+// Construct multisig signature using the address, the full set of all addresses, and the required signatures
+const multisigSig = await MultisigSignature.create(
+  multisigAddress, [bob.address, aliceAddress], signatures,
+)
+
+// Update signature on payment trasnaction
+paymentTxn.signature = multisigSig.bin
```
allenan, line 256: it would be nice if we didn't have to mutate the payment txn object here. maybe we could add a `multiSign` method in addition to `sign`?

allenan: I wonder if there would be a generic enough way to do it where it could be added at the `Transaction` parent class and call the implementation of `sign`?

packages/crypto/src/KeySignature.ts
```diff
@@ -0,0 +1,25 @@
...
+  public index!: number
+
+  public signature!: Uint8Array
+
+  constructor(index: number, signature: Uint8Array) {
+    this.index = index
+    this.signature = signature
+  }
+
+  public static new(addresses: Address[], address: Address, signature: Uint8Array): KeySignature {
```
allenan, line 13: could this be handled in the constructor itself?

packages/crypto/src/MultisigSignature.ts
```diff
@@ -0,0 +1,94 @@
...
+  public addresses!: Address[]
+
+  public signatures!: KeySignature[]
+
+  constructor(addresses: Address[], signatures: KeySignature[]) {
+    this.addresses = addresses
+    this.signatures = signatures
+  }
+
+  public static create(
```
allenan, line 17: same comment here. If there's some validation for what comprises a "valid" instance of MultisigSignature, it probably belongs in the constructor. and I notice that the other static method in this class is referencing this one, so probably safe to move to constructor

allenan:
This is great work! my main feedback is that multisig feels a little shoehorned into the transactions library. It would be great to have a convenient way of calling it on an instance of a transaction, without requiring the instance to be mutated after the fact.

packages/crypto/src/KeySignature.ts
jcronyn: I'd personally leave them separate, I think this pattern allows `new` to be used in cases where you want to use the list of addresses to find the signature, but you may have an aggregator accepting serialized `KeySignature` objects (from `KeySignature(...).bin`) instead, in which case the constructor would be more appropriate to use.

packages/crypto/src/MultisigSignature.ts
jcronyn: Similar to my other response I think this allows different uses. 

The `create` method confirms you have the correct number of addresses and sufficient signatures. If instead you want to just create a `MultisigSignature` with a set list of Addresses and KeySignatures (perhaps because it's partially signed and you're forwarding the payload to the next signer, let's say), then you can use the constructor directly.

Let me know if this still doesn't align with your thinking and happy to discuss further.

README.md
jcronyn: @allenan - I updated `MultisigSignature` so that it matches the `SignableKeypair` interface, so now you can just call `transactionInstance.sign({payer:multisigSignature})`

Thanks for the suggestion, let me know if this doesn't work 👍

packages/crypto/src/KeySignature.ts
allenan: My concern with having them be separate is that it appears that there is essential validation happening in this static method for what comprises a valid instance of a KeySignature. It's idiomatic javascript to use the `new` keyword to instantiate a class. so my worry is that we're opening up the possibility of creating invalid instances if a user reaches for a `new KeySingature` before they see the `KeySignature.create` method.

packages/crypto/src/MultisigSignature.ts
allenan: similar feedback. if the `insufficient signatures` or `wrong number of addresses` validation applies to any instance of a MultisigSignature class, then I'd expect that to happen in the constructor directly. if that assumption is untrue, then maybe something about the method naming should change to not insinuate that `create` is the proper way to instantiate a properly constructed instance

packages/crypto/src/KeySignature.ts
jcronyn: I'm less familiar with idiomatic javascript, but that makes sense to me 👍  I'll update the method signatures and move things around. @allenan - Would you suggest eliminating the `create` method altogether?

allenan: if the only purpose of the create method is to instantiate a valid instance of the class, then yes, that's for the constructor to do. if it has another purpose, then maybe we can look at another name for it?

syuan100:
Hey y'all, wanted to check in on this for any outstanding items! Can we aggregate the last few asks if any and post it as a comment for visibility? Thanks y'all, excited to get this in 🍻

packages/crypto/src/MultisigSignature.ts
jcronyn: Updated per our conversation, please let me know if any further updates needed. 😄

allenan (approved):
this looks good to me @jcronyn. I'll look into the conflicts

PR #292, opened 2022-04-20
Added reward type to the Reward model

andreiguzga:
I was trying to move from the activity endpoint and use the new rewards endpoint but this was missing the type of the reward. I've noticed that the type of the reward was added in the API but it is not reflected in this library. 

I am working on updating my platform to be compatible with the new changes that the light hotspots will add to validators and I need to know the types of rewards when fetching them through this library.


andreiguzga:
Any chance someone can take a look at this PR? I would like to prepare our platform to be compatible with the new switch to light hotspots and we are blocked by not having the reward type on the rewards endpoint.

Thanks

PR #293, opened 2022-04-22
Feature/verify burn

matthewcarlreetz:


packages/transactions/package.json
```diff
@@ -30,6 +30,7 @@
     "@helium/address": "^4.0.0",
     "@helium/proto": "^1.5.0",
     "@types/libsodium-wrappers": "^0.7.8",
+    "libsodium-wrappers": "^0.7.10",
```
allenan, line 33: i think generally we had kept this library crypto-agnostic so that it wouldn't crash trying to load an incompatible dependency on react-native or nodejs. Maybe we could expose verification in the crypto libraries?

matthewcarlreetz: @allenan Makes sense, I had that thought too. I created a `message` function in place of verify and added a `verify` utility to the crypto packages. `message` is unimplemented in the other txn classes, I can add them at some point, but would like to come back to that.

allenan (approved):
looks good

PR #306, opened 2022-05-26
Create mobile balance.

matthewcarlreetz:


packages/currency/src/CurrencyType.ts
```diff
@@ -5,6 +5,7 @@ import {
   USDollars,
   TestNetworkTokens,
 } from './currency_types'
+import MobileTokens from './currency_types/MobileTokens'
 
 export type AnyCurrencyType =
```
tyler-whitman, line 10: We should add `MobileTokens` to `AnyCurrencyType`

tyler-whitman (approved):
Looks good, should we wait to merge? Also I assume accounts will have a mobileTokenBalance but that is probably TBD.

PR #311, opened 2022-06-30
Feature/wallet link

matthewcarlreetz:
Create wallet link package that handles linking helium wallets to 3rd party apps.


packages/wallet-link/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@helium/wallet-link",
+  "version": "4.4.0",
```
tyler-whitman, line 3: Should this be 1.0.0?

packages/wallet-link/src/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * [[include:crypto-react-native/README.md]]
+ * @packageDocumentation
+ * @module crypto-react-native
```
tyler-whitman, line 4: This module should probably be something like `@helium/wallet-link`?

packages/wallet-link/src/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * [[include:crypto-react-native/README.md]]
```
tyler-whitman, line 2: This path is also wrong I think?

packages/wallet-link/src/walletLink.ts
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable object-curly-newline */
+import Sodium from 'react-native-sodium'
+import Address from '@helium/address'
+import queryString from 'query-string'
+import { getUnixTime } from 'date-fns'
+import { Keypair } from '@helium/crypto-react-native'
```
allenan, line 6: importing keypair from crypto-react-native will limit this package to only work on react native. is there any chance we want this to work from web with `@helium/crypto`? given that we're expecting the implementer to pass in a keypair anyway, we could structure this package like we do with `@helium/transactions` where we don't import either of the crypto libraries directly, but just expect a "signable" keypair-like object to be passed in

tyler-whitman (approved):

packages/wallet-link/package.json
allenan: all the lerna package versions are linked to update together, so this would start at whatever the current version overall is

packages/wallet-link/src/walletLink.ts
matthewcarlreetz: Great suggestion. I was trying to figure out how to do this.

packages/wallet-link/package.json
matthewcarlreetz: I'm not sure. @allenan ?

matthewcarlreetz: I ran `learn create` and it generated this for me 🤷

tyler-whitman: Ah ok no problem then

packages/wallet-link/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "access": "public"
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1",
+    "clean": "rimraf build",
+    "build": "yarn run clean && tsc"
+  },
+  "dependencies": {
+    "@helium/address": "^4.3.1",
+    "@helium/crypto-react-native": "^4.3.1",
```
tyler-whitman, line 30: We should probably remove this now and remove the usage in the test?

packages/wallet-link/src/walletLink.ts
tyler-whitman: It looks like the change you made isnt working properly. When I try to run `yarn docs` I get a bunch of type errors about `Type 'Keypair' is not assignable to type 'SignableKeypair'`

packages/wallet-link/src/walletLink.ts
```diff
@@ -0,0 +1,117 @@
+/* eslint-disable object-curly-newline */
+import Address from '@helium/address'
+import { utils } from '@helium/crypto'
```
matthewcarlreetz, line 3: @tyler-whitman how does this get swapped for `@helium/crypto-react-native` on react native? I'm getting an error when using it in the iOS simulator
Simulator Screen Shot - iPhone 13 - 2022-07-01 at 08 38 22

allenan: yeah you can't import either without causing an error on the wrong platform

allenan: maybe look at how `@helium/transactions` does its tests? you could probably make a mock signable keypair object for tests. you could also try moving crypto in dev dependencies but that may still get imported if the RN app is being run in dev mode

packages/wallet-link/package.json
Perronef5, line 3: Should this package not start at `1.0.0` or do we keep all these packages in sync?

Perronef5 (approved):

PR #325, opened 2022-07-25
fix token type

allenan:
token type comes from the API as a lowercase ticker. We will preserve that interface up until transaction serialization at which point it gets converted to an integer, following the protobuf enum. This also removes the dependency on `@helium/transactions` from `@helium/currency` and `@helium/http`

### DIFF
--- a/packages/http/src/models/Transaction.ts
+++ b/packages/http/src/models/Transaction.ts
@@ -11,7 +11,6 @@ import {
   IotTokens,
   MobileTokens,
 } from '@helium/currency'
-import { TokenType } from '@helium/transactions'
 import Challenge, { HTTPChallengeObject } from './Challenge'
 import DataModel from './DataModel'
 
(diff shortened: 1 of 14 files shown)


packages/http/src/models/Transaction.ts
```diff
@@ -495,7 +494,11 @@ function prepareTxn(txn: any, { deep } = { deep: false }) {
 
   balanceConversions.forEach(({ key, currencyType }) => {
     if (txn[key] && typeof txn[key] === 'number') {
-      txn[key] = new Balance(txn[key], currencyType)
+      if (txn.token_type !== undefined) {
```
matthewcarlreetz, line 497: also wondering if we should handle testnet here too

matthewcarlreetz: I'm not sure if `token_type` for testnet will be `'tnt'` or `'hnt'`, but that would affect how we handle testnet

matthewcarlreetz (commented):
Looking good, I think it would be valuable to handle testnet with this PR as well.

allenan:
> Looking good, I think it would be valuable to handle testnet with this PR as well.

sounds like testnet does not alter the token_type fields in its responses, so any handling we want to do for testnet will need to happen outside of the currency/token_type level

matthewcarlreetz (approved):

PR #327, opened 2022-08-01
Fix hnt payments failing to verify.

matthewcarlreetz:


packages/transactions/src/__tests__/PaymentV2.spec.ts
```diff
@@ -115,14 +126,67 @@ describe('sign', () => {
...
+  })
+
+  it('should decode hnt token type to hnt', async () => {
+    const payment = await paymentFixture()
+    const { bob } = await usersFixture()
+    const signedPayment = await payment.sign({ payer: bob })
+    const decoded = proto.helium.blockchain_txn.decode(
+      Buffer.from(signedPayment.toString(), 'base64'),
+    )
+    expect(decoded.paymentV2?.payments?.[0].tokenType).toEqual(TokenType.hnt)
```
matthewcarlreetz, line 180: When we create the payment `token_type` needs to be undefined, but when it's decoded it is 0. 🤔

allenan: do we need the same kind of logic in the `fromString` method?

allenan: hm I see this is calling the proto def directly

allenan (approved):

PR #329, opened 2022-08-02
Add default formatting options to balance

matthewcarlreetz:
Creates formatting defaults for currencies when using `toString` 
Allows overriding defaults `balance.toString(10, { showTrailingZeroes: false })`

The only currency specifying default formatting is USD which is...
`{ decimalPlaces: 2, showTrailingZeroes: true }`

### DIFF
--- a/packages/currency/src/Balance.ts
+++ b/packages/currency/src/Balance.ts
@@ -40,13 +40,6 @@ type StringFormatOptions = {
 
 const DC_TO_USD_MULTIPLIER = 0.00001
 
-const getVal = <T>(localValue?: T, defaultValue?: T) => {
-  if (localValue !== undefined) {
-    return localValue
-  }
-  return defaultValue
-}
-
 export default class Balance<T extends BaseCurrencyType> {
   public type: T
 
@@ -78,11 +71,8 @@ export default class Balance<T extends BaseCurrencyType> {
     const showTicker = options?.showTicker === undefined ? true : options.showTicker
     const format = { decimalSeparator, groupSeparator, groupSize: 3 }
     const roundingMode = options?.roundingMode || BigNumber.ROUND_DOWN
-    const decimalPlacesToDisplay = getVal(maxDecimalPlaces, this.type.format?.decimalPlaces)
-    const keepTrailingZeroes = getVal(
-      options?.showTrailingZeroes,
-      this.type.format?.showTrailingZeroes,
-    )
+    const decimalPlacesToDisplay = maxDecimalPlaces ?? this.type.format?.decimalPlaces
+    const keepTrailingZeroes = options?.showTrailingZeroes ?? this.type.format?.showTrailingZeroes
 
     let numberString = ''
     if (decimalPlacesToDisplay !== undefined && decimalPlacesToDisplay !== null) {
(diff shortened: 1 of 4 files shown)


packages/currency/src/Balance.ts
```diff
@@ -35,10 +35,18 @@ type StringFormatOptions = {
...
+  showTrailingZeroes?: boolean
 }
 
 const DC_TO_USD_MULTIPLIER = 0.00001
 
+const getVal = <T>(localValue?: T, defaultValue?: T) => {
+  if (localValue !== undefined) {
+    return localValue
+  }
+  return defaultValue
```
Perronef5, line 47:
```suggestion
  return localValue ?? defaultValue
```

packages/currency/src/Balance.ts
```diff
@@ -69,19 +77,29 @@ export default class Balance<T extends BaseCurrencyType> {
...
+    )
 
     let numberString = ''
-    if (maxDecimalPlaces !== undefined && maxDecimalPlaces !== null) {
-      numberString = this.bigBalance.toFormat(
-        maxDecimalPlaces,
-        options?.roundingMode || BigNumber.ROUND_DOWN,
-        format,
-      )
+    if (decimalPlacesToDisplay !== undefined && decimalPlacesToDisplay !== null) {
```
Perronef5, line 88:
```suggestion
    if (decimalPlacesToDisplay) {
```

packages/currency/src/Balance.ts
```diff
@@ -69,19 +77,29 @@ export default class Balance<T extends BaseCurrencyType> {
...
+        )
+      }
+      numberString = this.bigBalance.toFormat(decimalPlaces, roundingMode, format)
     } else {
       numberString = this.bigBalance.toFormat(format)
     }
-    // if it's an integer, just show the integer
-    if (parseInt(numberString.split(decimalSeparator)[1], 10) === 0) {
+    // if it's an integer and keepTrailingZeroes is false, just show the integer
+    if (!keepTrailingZeroes && parseInt(numberString.split(decimalSeparator)[1], 10) === 0) {
```
Perronef5, line 102: This could be better in case there is no decimal for some reason
```suggestion
    if (!keepTrailingZeroes) {
```

Perronef5 (approved):
LGTM! Just some minor comments

packages/currency/src/Balance.ts
allenan: I was trying to remember this operator recently and couldn't remember what to google, much nicer than checking undefined

packages/currency/src/Balance.ts
matthewcarlreetz: This doesn't work because I need it to go into that if statement if `decimalPlacesToDisplay === 0`

packages/currency/src/Balance.ts
matthewcarlreetz: Not sure I follow... I actually think that whole if block can be removed now.

packages/currency/src/Balance.ts
matthewcarlreetz: @allenan @Perronef5 There was an issue with app center not liking it, so I put it out of my mind, but I believe it is fine now.

packages/currency/src/Balance.ts
Perronef5: That too. I was just saying we don't need to check if there is a trailing zero because the split would work anyway

allenan (approved):

PR #330, opened 2022-08-03
Add max pay to PaymentV2 payments.

matthewcarlreetz:


packages/transactions/src/PaymentV2.ts
```diff
@@ -20,11 +20,12 @@ interface PaymentOptions {
   signature?: Uint8Array
 }
 
-interface Payment {
+export interface Payment {
   payee: Addressable
   amount: number
   memo?: Base64Memo
   tokenType?: string
+  max?: boolean | null
```
tyler-whitman, line 28: Do we need the `| null` with the `?`

tyler-whitman (approved):

packages/transactions/src/PaymentV2.ts
matthewcarlreetz: That's how the proto is defined, but I suppose it would be better to keep it more in line with how this package handles that.

allenan (approved):
looks good 👍

PR #338, opened 2022-10-25
Add sol token currency type.

matthewcarlreetz:


packages/currency/src/Balance.ts
```diff
@@ -117,7 +118,11 @@ export default class Balance<T extends BaseCurrencyType> {
   toNetworkTokens(oraclePrice?: Balance<USDollars>): Balance<NetworkTokens> {
     if (this.type instanceof NetworkTokens) return this
 
-    if (this.type instanceof MobileTokens || this.type instanceof IotTokens) {
+    if (
+      this.type instanceof MobileTokens ||
+      this.type instanceof IotTokens ||
+      this.type instanceof SolTokens
```
allenan, line 124: do we also need to mark `toUsd` and `toDataCredits` as not a supported conversion direction from sol?

matthewcarlreetz: For both of those cases `SolTokens` won't match any of the `if` statements, so it will fall through to the default case `throw UnsupportedCurrencyConversionError`

allenan (approved):

PR #342, opened 2022-10-28
Create balance from ticker.

matthewcarlreetz:


packages/currency/src/Balance.ts
```diff
@@ -66,6 +67,16 @@ export default class Balance<T extends BaseCurrencyType> {
     return new Balance(integerBalance, currencyType)
   }
 
+  static fromIntAndTicker(integerBalance: number, ticker?: Ticker) {
+    const currencyType = CurrencyType.fromTicker(ticker)
```
allenan, line 71: ticker is just a string right? what if this fails to find a matching currency type? how does that get handled?

packages/currency/src/currency_types/index.ts
```diff
@@ -7,3 +7,5 @@ export { default as BaseCurrencyType } from './BaseCurrencyType'
 export { default as MobileTokens } from './MobileTokens'
 export { default as SolTokens } from './SolTokens'
 export { default as IotTokens } from './IotTokens'
+import { Ticker } from './BaseCurrencyType'
```
allenan, line 10: do you need to import and then export? or could you combine into an `export ... from` statement?

matthewcarlreetz: yeah, I don't know why, but that wasn't working for me last night, but it is today. lol.

allenan (approved):

PR #357, opened 2023-02-17
Change IOT and MOBILE decimal to 6

Perronef5:
## TODO 📝
- [x] Add fetch by mint option for currencies

### DIFF
--- a/packages/currency/package.json
+++ b/packages/currency/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@helium/currency",
-  "version": "5.0.0",
+  "version": "4.9.0",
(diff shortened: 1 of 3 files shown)


packages/currency/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@helium/currency",
-  "version": "4.9.0",
+  "version": "5.0.0",
```
tyler-whitman, line 3: You dont need to change this manually, When we run a release it will auto update. It should be reverted

tyler-whitman (approved):

PR #363, opened 2023-09-07
Feature/mobile wifi onboarding

matthewcarlreetz:
Assertion fee estimation is working well in the wallet app. 

The wifi onboarding is just a first pass. It will likely need some refactoring once it is used in an app.


matthewcarlreetz (commented):
Had to upgrade typescript which caused some errors in the http library. I added some ts-ignores. Alternatively, we could just remove the package now?

packages/onboarding/src/getAssertData.ts
```diff
@@ -0,0 +1,519 @@
...
+  mobileInfoKey,
+  rewardableEntityConfigKey,
+} from '@helium/helium-entity-manager-sdk'
+import { init as initHem } from '@helium/helium-entity-manager-sdk'
+import * as Currency from '@helium/currency-utils'
+import { HotspotType } from './types'
+
+export const TXN_FEE_IN_LAMPORTS = 5000
+export const TXN_FEE_IN_SOL = TXN_FEE_IN_LAMPORTS / LAMPORTS_PER_SOL
+export const FULL_LOCATION_STAKING_FEE = 1000000 // $10 - does this need to be updated to $5? It's used as a fallback when something fails
```
ChewingGlass, line 35: This can be fetched from the value on chain. You need to fetch the `RewardableEntityConfig`

packages/onboarding/src/getAssertData.ts
```diff
@@ -0,0 +1,519 @@
...
+  {
+    maker,
+    connection,
+    owner,
+    rpcEndpoint,
+  }: { maker: PublicKey; connection: Connection; owner: PublicKey; rpcEndpoint: string },
+) => {
+  const walletDC = await getAssociatedTokenAddress(new PublicKey(DC_MINT), owner)
+
+  const makerDC = await getAssociatedTokenAddress(new PublicKey(DC_MINT), maker)
```
ChewingGlass, line 273:
```suggestion
  const makerDC = getAssociatedTokenAddressSync(new PublicKey(DC_MINT), maker)
```

packages/onboarding/src/getAssertData.ts
```diff
@@ -0,0 +1,519 @@
...
+const estimateMetaTxnFees = async (
+  buff: Buffer,
+  {
+    maker,
+    connection,
+    owner,
+    rpcEndpoint,
+  }: { maker: PublicKey; connection: Connection; owner: PublicKey; rpcEndpoint: string },
+) => {
+  const walletDC = await getAssociatedTokenAddress(new PublicKey(DC_MINT), owner)
```
ChewingGlass, line 271:
```suggestion
  const walletDC = getAssociatedTokenAddressSync(new PublicKey(DC_MINT), owner)
```

packages/onboarding/src/getAssertData.ts
```diff
@@ -0,0 +1,519 @@
...
+        ownerFees: {
+          lamports: number
+          dc: number
+        }
+        isFree: boolean
+      }
+    | undefined
+  )[]
+  try {
+    simulatedFees = await Promise.all(
```
ChewingGlass, line 378: I don't think this works in the case that the maker wallet has insufficient funds. The simulation will probably fail, and then it will put all fees on the owner. But the onboarding server does not have logic to change the payer, maker is still the payer not the user.

ChewingGlass:
A lot of this logic is duplicated from the react native sdk. I think we need to clean up this code, remove the Balance stuff, and then import it back to react native helium, removing the duplicate code, and do a major release. 

Otherwise we're going to have to maintain all this crazy fee estimation onboarding logic in two places.

ChewingGlass (changes requested):

matthewcarlreetz:
> A lot of this logic is duplicated from the react native sdk. I think we need to clean up this code, remove the Balance stuff, and then import it back to react native helium, removing the duplicate code, and do a major release.
> 
> Otherwise we're going to have to maintain all this crazy fee estimation onboarding logic in two places.

@ChewingGlass I totally agree. My intent is to remove the duplicated logic from the React Native SDK once this goes live. I'll work through your comments now.

packages/onboarding/src/SolanaOnboarding.ts
```diff
@@ -0,0 +1,188 @@
...
+      elevation,
+      maker,
+      nextLocation: location,
+      hotspotTypes,
+      cluster: this.cluster,
+    })
+  }
+
+  keyToAsset = async (hotspotAddress: string) => {
+    if (!this.shouldMock) {
```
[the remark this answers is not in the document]
matthewcarlreetz: yeah, agreed

packages/onboarding/src/WifiHttpClient.ts
```diff
@@ -0,0 +1,86 @@
...
+      },
+    })
+
+    if (mockRequests) {
+      this.mockAdapater = new MockAdapter(this.axios, { delayResponse: 1000 })
+      this.mockAdapater.onPost('/on_hotspot_nft_created').reply(200, {})
+    }
+  }
+
+  getTxnFromGateway = async (payerHeliumAddress: string) => {
```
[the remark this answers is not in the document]
matthewcarlreetz: I generally don't unless TS can't infer them for some reason.

packages/onboarding/src/getAssertData.ts
matthewcarlreetz: yeah, I have that happening already. This is just a fallback in case the lookup fails.

packages/onboarding/src/getAssertData.ts
matthewcarlreetz: @ChewingGlass We have this logic throwing an error in the simulation.
```ts
  if (response.data.result.value.err) {
    console.error(response.data.result.value.logs.join('\n'))
    throw new Error('Transaction would fail')
  }
  ```
  Is there any way to detect that it's the maker who doesn't have the funds here? Or would it be best to throw a specific error from the `update-metadata` onboarding server request?

File: packages/onboarding/src/AssertUtil.ts [hunk 7]
```diff
@@ -0,0 +1,488 @@
...
+}
+
+const getAtaAccountCreationFee = async (solanaAddress: PublicKey, connection: Connection) => {
+  const ataAddress = getAssociatedTokenAddressSync(DC_MINT, solanaAddress, true)
+
+  try {
+    await getAccount(connection, ataAddress)
+    return new BN(0)
+  } catch {
+    return new BN(2039280)
```
ChewingGlass, line 109: Should use the connection to get the rent cost for the size of the account, as rent may be dynamic in the near to mid future.

packages/onboarding/src/AssertUtil.ts
```diff
@@ -0,0 +1,488 @@
...
+    configKey[0],
+  )
+  const config = isIOT ? entityConfig?.settings.iotConfig : entityConfig?.settings.mobileConfig
+
+  // @ts-ignore
+  const configFee = config?.fullLocationStakingFee as BN
+  if (configFee) {
+    return toBN(configFee, 0)
+  }
+  return new BN(FULL_LOCATION_STAKING_FEE)
```
ChewingGlass, line 130: This will no longer work. https://github.com/helium/wallet-app/blob/main/src/hooks/useOnboardingBalances.ts#L141

packages/onboarding/src/AssertUtil.ts
```diff
@@ -0,0 +1,488 @@
...
+const getBalance = async (wallet: PublicKey, connection: Connection, mint: PublicKey) => {
+  const balance = await Currency.getBalance({
+    pubKey: wallet,
+    connection,
+    mint: new PublicKey(mint),
+  })
+  return new BN(balance.toString())
+}
+
+const getBalances = async (wallet: PublicKey, connection: Connection) => {
```
ChewingGlass, line 55: Trying to avoid @helium/currency

packages/onboarding/src/AssertUtil.ts
```diff
@@ -0,0 +1,488 @@
...
+  maker,
+  dcMint,
+}: {
+  owner: { key: PublicKey; account: Account; dcAccount: Account }
+  maker: { key: PublicKey; account: Account; dcAccount: Account }
+  connection: Connection
+  dcMint: PublicKey
+}) => {
+  const makerFees = await getAccountFees({ connection, ...maker, dcMint })
+  const ownerFees = await getAccountFees({ connection, ...owner, dcMint })
```
ChewingGlass, line 223: I think doing something more like this: https://github.com/helium/wallet-app/blob/main/src/hooks/useOnboardingBalances.ts#L141

Is a better approach than simulating for fees. That way you can see if balances are proper before creating the transaction.

packages/onboarding/src/OnboardingClient.ts
```diff
@@ -156,12 +207,40 @@ export default class OnboardingClient {
...
+    deviceType,
+  }: {
+    onboardingKey: string
+    authToken: string
+    deviceType: DeviceType
+  }) {
+    return this.axios.post(
+      'hotspots',
+      {
+        batch: 'example-batch',
```
ChewingGlass, line 231: Should be a param, right? Same with the serial, eth0, etc.

packages/onboarding/src/SolanaOnboarding.ts
```diff
@@ -0,0 +1,292 @@
...
+    return createTransferCompressedCollectableTxn({
+      collectable: asset,
+      owner: this.wallet,
+      recipient: new PublicKey(newOwner),
+      connection: this.connection,
+      url: rpcEndpoint,
+    })
+  }
+
+  createTransferInstructions = async ({
```
ChewingGlass, line 266: Does this need to be part of onboarding?

ChewingGlass (changes requested):

packages/onboarding/src/HotspotOnboardingUtil.ts
```diff
@@ -0,0 +1,440 @@
...
+
+    const [info] = mobileInfoKey(configKey[0], address)
+
+    const mobileInfo = await hemProgram.account.mobileHotspotInfoV0.fetch(info)
+    const deviceTypeKeys = Object.keys(mobileInfo.deviceType)
+    let deviceType = ''
+    if (deviceTypeKeys.length) {
+      deviceType = deviceTypeKeys[0]
+    }
+    const mobileConfigV1 = entityConfig.settings.mobileConfigV1
```
ChewingGlass, line 134: HIP-96 is forcing us toadd a mobileConfigV2... https://github.com/helium/helium-program-library/pull/506. Can you add support for this if present?

packages/onboarding/src/HotspotOnboardingUtil.ts
```diff
@@ -0,0 +1,440 @@
...
+    if (!deviceFees) return details
+    return {
+      ...details,
+      locationStakingFee: deviceFees.locationStakingFee,
+    }
+  } catch (e) {
+    const error = String(e)
+    if (error.startsWith('Error: Account does not exist or has no data')) {
+      // This is an expected error. It has not been onboarded to this network
+      return
```
ChewingGlass, line 151: If you use .fetchNullable instead of fetch, you won't get this error

packages/onboarding/src/HotspotOnboardingUtil.ts
```diff
@@ -0,0 +1,440 @@
...
+      // @ts-ignore
+      (fee) => fee.deviceType[lowerFirst(deviceType)] !== undefined,
+    )
+
+    // @ts-ignore
+    const details = hotspotInfoToDetails(mobileInfo)
+    if (!deviceFees) return details
+    return {
+      ...details,
+      locationStakingFee: deviceFees.locationStakingFee,
```
ChewingGlass, line 145: You might as well also return the onboarding fees.

ChewingGlass (changes requested):

ChewingGlass (approved):

PR #368, opened 2024-03-08
onboarding rando

matthewcarlreetz:
- Add more logs, cleanup unused funcs.
- Verify config message before sending.
- Throw an error when firmware lookup fails to respond.
- Cleanup incorrect error logging when onboarding
- Cleanup logs.
- Add cbrs suppot to mobile onboarding client.
- Use config api dev url when onboarding to devnet.
- Add hotspot to onboarding server after retrieving add_gateway_txn

### DIFF
--- a/packages/onboarding/src/HotspotOnboardingUtil.ts
+++ b/packages/onboarding/src/HotspotOnboardingUtil.ts
@@ -8,10 +8,8 @@ import {
   heliumAddressToSolPublicKey,
 } from '@helium/spl-utils'
 import { subDaoKey } from '@helium/helium-sub-daos-sdk'
-import { Connection, PublicKey, AccountInfo, Cluster, LAMPORTS_PER_SOL } from '@solana/web3.js'
-import axios from 'axios'
+import { Connection, PublicKey, Cluster, LAMPORTS_PER_SOL } from '@solana/web3.js'
 import {
-  AccountLayout,
   createAssociatedTokenAccountIdempotentInstruction,
   getAccount,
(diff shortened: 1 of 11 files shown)


packages/onboarding/src/HotspotOnboardingUtil.ts
```diff
@@ -402,8 +328,8 @@ export const getAssertData = async ({
 
     const dcInDollars = dcNeeded.div(new BN(100000))
     const oraclePrice = await getOraclePriceFromSolana({ connection, cluster })
-    const hntNeeded = dcInDollars.div(oraclePrice)
-    hasSufficientHnt = balances.hnt.gte(hntNeeded)
+    const hntNeeded = dcInDollars.toNumber() / oraclePrice.toNumber()
+    hasSufficientHnt = balances.hnt.toNumber() > hntNeeded
```
ChewingGlass, line 332: This will explode if for some reason they have more hnt than javadcript supports

ChewingGlass (changes requested):

packages/onboarding/src/HotspotOnboardingUtil.ts
matthewcarlreetz: @ChewingGlass updated. Can you take another look?

ChewingGlass (approved):

ChewingGlass (changes requested):
Can we call it hnt and not bones. Bones is overloaded, we use that for mobile/iot

ChewingGlass (approved):

